PR #1, opened 2020-05-13
Minor Build Generalization

garrettwrong:
- Swapped out some of the localized lines in the makefile with more common defaults.  This helps me to compile right out of the box on several machines.
- Adds CUB as a submodule.  That default submodule dir is the default location in makefile.  That should "just work" for most users if they use the provided clone line.
- Changed the NVCC arch targets to be much more general (compile is slower, but developer-users can trim in their own make.inc).
- Adds a make.inc.power9
- Changes to README to match above. Also changed perspective of the authorship line.

Because this is slightly different than the changes in my fork, I want to review with fresh eyes, and retest before un-drafting.  At that point I might suggest an independent person try it to get their install experience.  Good time for any discussion.


garrettwrong:
I had pretty good luck with these makefile changes on tigerGPU, Traverse (IBM) and my desktop today.  If anyone else would like to check out this branch and try to follow it's readme to confirm, that would be helpful.

MelodyShih:
I tried the new makefile and met no problem building the code on Courant's gpu cluster :) The commit is because the content in make.inc file looked out of date with the changes in the makefile.

ahbarnett:
I hope I'm not the merger here - I'm not set up to test on GPU, so someone else should try it.

janden:
> I hope I'm not the merger here - I'm not set up to test on GPU, so someone else should try it.

I'm going to review it now and test it out locally, don't worry!

janden:
So this didn't work for me out of the box. The issue was that the default CUDA install on Ubuntu doesn't include the code samples. I've added these in a `contrib` subdirectory that can be included. So far I only made this change for `make.inc.ubuntu`, but it may make sense to make this the default in the makefile (for consistency between builds).

Also, can we rename `makefile` to `Makefile`?

garrettwrong:
> So this didn't work for me out of the box. The issue was that the default CUDA install on Ubuntu doesn't include the code samples. I've added these in a `contrib` subdirectory that can be included. So far I only made this change for `make.inc.ubuntu`, but it may make sense to make this the default in the makefile (for consistency between builds).
> 
Interesting. I do always install the samples in the default dir within `/usr/local/cuda`, when using the CUDA installer. I guess the Princeton systems people did too.  I'm glad you caught this.  Thank you.  If it is just the two files, contrib should be fine. I can probably totally cull the sample location dependence.  Only if significant changes happen in CUDA would you replace those util functions.

I'll patch the makefile.

> Also, can we rename `makefile` to `Makefile`?
Sure.

I'll have this done shortly.

garrettwrong:
@janden , if you could take a another pass with the patch on your ubuntu machine, that would be helpful. Having this in master will simplify the procedure for hackathon, which I am writing up now.

janden:
> if you could take a another pass with the patch on your ubuntu machine, that would be helpful.

This works well on my machine now, no make.inc needed. Same goes for running on the FI cluster, where CUDA is loaded into the relevant paths using modules.

One thing to note is that the NVARCH setting will only work for CUDA>=10, although the code is compatible with CUDA 8 and 9. I assume there's no flexible way of specifying this (also not sure whether we're interested in supporting CUDA 8 and 9).

garrettwrong:
Cool, thanks, I'll make those changes and merge it in.  

Regarding the arch, yes this expects the nvidia toolchain is current (out over a year...), but generates code for all cards that toolchain supports (goes back 7-8 years).

The easiest portable thing is just to remove the optimization(s). You can also query which CUDA is available etc in the build. If it actually comes up it can be addressed then. I don't think it is a problem for anything relating to ASPIRE.

PR #3, opened 2020-05-18
nufft_opt as a pointer

garrettwrong:
Before I do spread_opts I wanted to run this by you guys in case we don't like it..

So I converted `cufinufft_plan` struct to have a pointer to nufft_opts. Then I converted all the code I found so far to use that.  It was in a lot of places... at least two cups of coffee worth.  Anyway, this is about 1/2 the change-set for converting to pointers.

I also don't know how this relates to the FINUFFT code changes that were mentioned...

If we would like to complete the same for shopts I'd be more than happy to.  Otherwise I'll stash this and clean up the python bindings/packaging.  Just let me know, thanks.

(note, this is branched off `minor_build_generalization`, would be rebased later obviously..

### DIFF
--- a/finufft/common.cpp
+++ b/finufft/common.cpp
@@ -28,30 +28,30 @@ void finufft_default_opts(nufft_opts &o)
 	o.modeord = 0;
 }
 
-int setup_spreader_for_nufft(spread_opts &spopts, FLT eps, nufft_opts opts)
+int setup_spreader_for_nufft(spread_opts &spopts, FLT eps, nufft_opts* opts)
 // Set up the spreader parameters given eps, and pass across various nufft
 // options. Report status of setup_spreader.  Barnett 10/30/17
 {
-  int ier=setup_spreader(spopts, eps, opts.upsampfac, opts.spread_kerevalmeth);
-  spopts.debug = opts.spread_debug;
-  spopts.sort = opts.spread_sort;     // could make dim or CPU choices here?
-  spopts.kerpad = opts.spread_kerpad; // (only applies to kerevalmeth=0)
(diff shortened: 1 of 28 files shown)


ahbarnett:
This looks fine to me, if painful to do. The compiler should indicate any missed ones! I noticed in our new finufft we are building we don't use pointers to structs - maybe we should too to aid the py interface there? (but we use pybind11).

garrettwrong:
It is really up to you guys re the structs.

There is a similar python binding package, `cffi`, for C bindings. I am going to attempt that on this code and push it as well, as a different draft.  It tries to be more clever than ctypes and is the preferred way to package python C extensions afaik (works with setup.py).  It will have its own "gotchas" (it needs pure C headers, so we'll need to preprocess them twice, once float, once double).

You might find the cffi is much closer to what you're doing with pybind11. So to answer your question, I'd say let's take a look at cffi first before you consider changing your data structures. If you like it, ctypes (and the struct concern) mostly vanishes.  I'll push it like this, enough that we can see what it looks like before polishing...

garrettwrong:
Actually, I ran into similar problems with cffi.  The headers are mixing C++ (in several includes) and would require a fair amount more work to be automatically parsed correctly as straight C.  It can be done, but looking at it so far, I don't think it would be an improvement. 

You may find pybind11 is better suited to this project if there isn't going to be a strict C api (there may not be if we're mixing in finufft and other stuff).

For now I will separate Joakim's python into a low level binding file and python api file, then complete the bindings for singles doubles using the existing ctypes. The low level bindings can always be swapped out later.  I''ll close this soon, but leave it in the tree for a while.

PR #4, opened 2020-05-19
FINUFFT dependency clean-up

MelodyShih:
Here are the summary of what I have changed:
- Functions and files from FINUFFT that are not used by cuFINUFFT are removed.
- Variables in `nufft_opts` and `spread_opts` that are not used by cuFINUFFT are deleted.
- `nufft_opts` struct is moved to cufinufft.h.
- `finufft` directory is renamed to `contrib`

p.s. This pull request may simplify the change of having options in `cufinufft_plan` as pointers.


ahbarnett:
This looks fine to me, if painful to do. The compiler should indicate any missed ones! I noticed in our new finufft we are building we don't use pointers to structs - maybe we should too to aid the py interface there? (but we use pybind11).

MelodyShih:
To move all the header files into include/ directory, some files under contrib/ will also need to be moved. Do you have any suggestions regarding this?

ahbarnett:
GOod point. We actually left legendre_rule_fast.h in contrib. You could leave all unmodified FINUFFT headers in contrib too, unless the GPU user needs to see them.  The point is that include/ is the standard place for "user-facing" .h to be.  This is not very important.

janden:
> The point is that include/ is the standard place for "user-facing" .h to be. 

Yeah I think this makes sense. The header files in `contrib/` are only accessed internally by the code, right? Ideally, we'd only have one file in `include/`, which would be `cufinufft.h`.

Makefile
```diff
@@ -55,10 +58,10 @@ DYNAMICCLIB=lib/$(CLIBNAME).so
 
 HEADERS = src/cufinufft.h src/deconvolve.h src/memtrasfer.h src/profile.h \
 	src/spreadinterp.h
-FINUFFTOBJS=finufft/utils.o finufft/dirft2d.o finufft/common.o \
-	finufft/spreadinterp.o finufft/contrib/legendre_rule_fast.o
+CONTRIBOBJS=contrib/utils.o contrib/dirft2d.o contrib/common.o \
+	contrib/spreadinterp.o contrib/legendre_rule_fast.o
 
-CUFINUFFTOBJS=src/2d/spreadinterp2d.o src/2d/cufinufft2d.o \
+CUCONTRIBOBJS=src/2d/spreadinterp2d.o src/2d/cufinufft2d.o \
```
janden, line 64: These should still be `CUFINUFFTOBJS`, no?

contrib/removed/finufft.h
```diff
@@ -0,0 +1,22 @@
+#ifndef FINUFFT_H
```
janden, line 1: Why are we keeping this file?

include/cudeconvolve.h
```diff
@@ -1,7 +1,7 @@
-#ifndef __DECONVOLVE_H__
```
janden, line 1: This header isn't part of the user-facing API, right? Maybe it can stay in `src/`? I don't know what a good convention is here.

include/cufinufft.h
```diff
@@ -5,13 +5,31 @@
...
-#include "../finufft/spreadinterp.h"
-#include "../finufft/finufft.h"
-#include "../finufft/common.h"
+#include "../contrib/utils.h"
+#include "../contrib/spreadinterp.h"
+#include "../contrib/common.h"
 
 enum finufft_type {type1,type2,type3};
 
+struct nufft_opts {   // see cufinufft_default_opts() for defaults
```
janden, line 14: Since these options are specific to cufinufft, should it perhaps be called `cufinufft_opts`? Also, doesn't this need a `typedef`?

include/cuspreadinterp.h
```diff
@@ -1,8 +1,7 @@
+#ifndef __CUSPREADINTERP_H__
```
janden, line 1: Again, this could probably stay in `src/`.

include/memtransfer.h
```diff
@@ -1,7 +1,7 @@
 #ifndef __MEMTRANSFER_H__
```
janden, line 1: Also could stay in `src/`.

src/2d/spreadinterp2d.cu
```diff
@@ -57,7 +57,7 @@ void eval_kernel_vec_Horner(FLT *ker, const FLT x, const int w,
 	FLT z = 2*x + w - 1.0;         // scale so local grid offset z in [-1,1]
 	// insert the auto-generated code which expects z, w args, writes to ker...
 	if (upsampfac==2.0) {     // floating point equality is fine here
-#include "../../finufft/ker_horner_allw_loop.c"
+#include "../../contrib/ker_horner_allw_loop.c"
```
janden, line 60: Can this move up top with the other includes?

src/3d/spreadinterp3d.cu
```diff
@@ -52,7 +52,7 @@ void eval_kernel_vec_Horner(FLT *ker, const FLT x, const int w,
 	FLT z = 2*x + w - 1.0;         // scale so local grid offset z in [-1,1]
 	// insert the auto-generated code which expects z, w args, writes to ker...
 	if (upsampfac==2.0) {     // floating point equality is fine here
-#include "../../finufft/ker_horner_allw_loop.c"
+#include "../../contrib/ker_horner_allw_loop.c"
```
janden, line 55: Move up?

Makefile
MelodyShih: Yes, this was a mistake. Thanks for pointing this out.

contrib/removed/finufft.h
MelodyShih: This is again a mistake. Files under removed wasn't intended to be included....

include/cudeconvolve.h
MelodyShih: I agree (see commit 5ab18e4).

include/cuspreadinterp.h
MelodyShih: See 5ab18e4.

include/memtransfer.h
MelodyShih: See 5ab18e4.

include/cufinufft.h
MelodyShih: See 4f89eeb.

src/2d/spreadinterp2d.cu
MelodyShih: Since this file is from finufft, I am not sure if it is a good idea to put it at other place.

janden: I see. I didn't realize this was some sort of macro trick.

janden:
Looks good. I will merge this later today unless anyone objects.

ahbarnett:
Macro tricks is how I, and then Ludvig, made the kernel eval so fast...

src/cufinufft.cu
```diff
@@ -500,19 +499,19 @@ int cufinufft_default_opts(finufft_type type, int dim, nufft_opts &opts)
 		case 2:
 		{
 			opts.gpu_maxsubprobsize = 1024;
-			if(type == type1){
+			if(type == 1){
```
garrettwrong, line 502: Hey, if you remember could you tag me next time there is an API change like this?  1,2,3 is better, but I didn't realize it changed :) The enum it replaced started at 0. Naturally I thought I had a bug in my side of the code for a while :D..

MelodyShih: Yes, sure. Sorry for the trouble that it caused... I will make sure to let you know if similar changes made in the future.

janden: > Macro tricks is how I, and then Ludvig, made the kernel eval so fast...

Yeah I read that too quickly before I realized what was actually going on... Seems like a pretty neat solution.

janden:
> Macro tricks is how I, and then Ludvig, made the kernel eval so fast...

Yeah I read that too quickly before I realized what was actually going on... Seems like a pretty neat solution.

PR #6, opened 2020-05-19
add dual bindings in a wrapper class

garrettwrong:
@janden , something like this can complete the binding for both floating point types.  Python users would

```
nufft = cufinufft(dtype=np.float32)
...
nufft.plan(...)
nufft.execute(...)
nufft.destroy(...)
```

The code picks the correctly typed methods, structures and library. 

Its a little boiler plate.  At least the structs are not repeated. Obviously names easy to change etc.

If looks okay I can finish, clean, doc, and test it. For ASPIRE purposes, this plus some minimal setup.py packaging for a quasi release would be great short term goal.

One day if a better binding comes along, we just swap out the _back, hopefully little to no api change.


janden:
I think this would be fine, but would it be very different to specify the dtype at plan time? That way, the dtype decision is done locally and not at import time (I guess you could instantiate the nufft object locally each time, but it would amount to extra code).

garrettwrong:
I agree at plan instantiation would be better.  It was just a bit more code and I wanted to get a little feedback before I committed to completing. Thanks.

garrettwrong:
Hey, I have not forgotten about this. I got pretty much to where I was pre hackathon event today.  I have the dual typed bindings working locally, just getting all the unit tests to pass on our side (there was a lot of hard coded types I'm still sorting out).

I was able to make the init method cover plan more/less as Joakim suggested above, so it will instantiate a type bound plan..

Once I have that I'll commit the code over here.

garrettwrong:
This code seems to work well when in ASPIRE.  I have to think a little about how we are going to package/install it similar to finufft so ASPIRE can find the wrapper and libraries now that I moved them back to cufinufft. (for development I let them live in the ASPIRE dir..).

janden:
This all sounds good. Let me know when it's ready for review.

garrettwrong:
Actually, I have quick questions (for everybody)... while I'm making a minimal `setup.py` for other reasons..

- Who would you like to list as the Authors?
- email? I think the email must be singular. Was Alex in finufftpy.
- License? `Apache 2` same as finufftpy?
- Where would you like to start the versioning? (PyPI will want a version iirc).

@ahbarnett , @janden , @MelodyShih , thanks.

ahbarnett:
Authors: two-tier. Eg principal author: Melody. Main other contributors: other 3 of us.
Eg see here: https://finufft.readthedocs.io/en/latest/ackn.html
Will evolve over time.

email: Melody, since most likely to know details, or Garrett to first handle installation issues (I predict the main emails).

Yes, same as FINUFFT.
Flatiron prefers Apache 2, w/ copyright (C) year-range The Simons Foundation. But some work is now being done outside FI (and our IP policy allows author-ownership for those parts), so could add statement about that. I can email you our IP policy if you want. Or, J, check w/ Leslie.

version 1.0 beta as you get ready for release?

The above info should propagate the cufinufft as a whole; the py interface is just part.
Ie have it match whatever's in the README.md and the C++ code.

garrettwrong:
I wasn't clear, this is just for PyPI packaging, so we need to live in their constraints. We get one email, and a single list of authors. Sometimes people setup a project email, but that would be something at FI... 

I'm not worried about IP from my end. I'm work for hire :). Apache 2 it is.

garrettwrong:
Don't worry about all the commits, I'll probably squash this PR or at least rebase it :).

We have a pip installable module that is dtype aware.

It still depends on the CUDA library being both built and in the library path. To get away from that is a sort of big project that wouldn't add a lot of value to the main CUDA stuff. It is possible with CMake, but my experiences with it have had just as much or more complications, and would complicate the main project as well.

You may also run pytest and generate sphinx RTD style docs in the `cufinufftpy` directory now.

On the ASPIRE side I have mating code which expects this package and will try to use if possible.

janden:
> Don't worry about all the commits, I'll probably squash this PR or at least rebase it :).

All right, so is this ready for test run (if not a review)?

> It still depends on the CUDA library being both built and in the library path. 

I think this is fine for now.

> version 1.0 beta as you get ready for release?

Think this would be a little hasty. Maybe 0.1?

garrettwrong:
Yes, you can try to build it, build the python wrapper, and run pytest.  I tested on my development box, but I haven't tested on a totally vanilla machine (takes few minutes to set that up in the cloud, next steps).

Probably between your testing and the meeting later we can figure out if we need major re-thinking of the approach.

janden:
So I was able to get this to work on my laptop after some messing around. The main issue for me was that there's no easy way to make both double- and single-precision versions of the library through the makefile. What I essentially had to do was
```
make clib
mv lib/libcufinufftc.so .
make clean
PREC=SINGLE make clib
mv lib/libcufinufftcf.so .
```
since `make clean` will remove the first (double-precision) library and running `PREC=SINGLE make clib` *without* first running `make clean` will result in a corrupted library (since the majority of the source code is not recompiled given that their names are the same).

In the first instance, it might make sense to look at how to build both versions of the library without forcing `make clean`. In the long term, a more viable solution might be to include both single- and double-precision versions in the same library (as we discussed).

garrettwrong:
There is, it was in a thread here recently.

Make -B

Sorry on phone, otherwise I'd explain more. Next time I talk to you...

garrettwrong:
https://github.com/flatironinstitute/cufinufft/pull/31#issuecomment-643761401

ahbarnett:
Don't the function namespaces collide if you link to both libs, or not because it's C++? (Sorry if dumb question).
We have to solve this same problem in finufft...  but at least it could just be that collisions only need to be prevented in C and the language wrappers.

garrettwrong:
The functions are in different libraries, which I select based on the
dtype, and for the functions with 'f'suffixed names I define the correct
name in the wrapper class. So at runtime we're getting the right functions
from the right library afaik.

Generally yes, it is a problem...  This is just one way to get around it
based on how the low level codes were already implemented, for
consideration. It would only require the additional make line I linked to.

For extern C functions the CPP macro thing can solve that problem by
building one library... like we talked about yesterday, that's just
different work.

janden:
> Make -B

Ah sorry I missed that. Reading up on that now. So I'd do something like
```
make clib
PREC=SINGLE make -B clib
```
Still not the most user-friendly approach, but it'll do for now.

garrettwrong:
No it is not :). I sort of viewed the underlying libraries as a different issue. It would make sense to finalize that concern before polishing the higher layers tbh :/.

garrettwrong:
Half the API work depends directly on the library/build process, and decisions made for that (here and in finufft it seems).

There's not much to the high level API. Could add some more checking etc, but the point of the effort is to get something we can pip install in ASPIRE, not highly buff the API. Nobody besides me and you will probably mess with the python API for a while anyway.

janden:
Ok. Let's focus on the library then first before finishing this off.

Makefile
```diff
@@ -68,22 +68,24 @@ CUFINUFFTOBJS_64=src/2d/spreadinterp2d.o src/2d/cufinufft2d.o \
...
-	$(CC) -c $(CFLAGS) $(INC) $< -o $@
-%.o: %.cu $(HEADERS)
-	$(NVCC) --device-c -c $(NVCCFLAGS) $(INC) $< -o $@
 %_32.o: %.cpp $(HEADERS)
 	$(CXX) -DSINGLE -c $(CXXFLAGS) $(INC) $< -o $@
 %_32.o: %.c $(HEADERS)
 	$(CC) -DSINGLE -c $(CFLAGS) $(INC) $< -o $@
 %_32.o: %.cu $(HEADERS)
 	$(NVCC) -DSINGLE --device-c -c $(NVCCFLAGS) $(INC) $< -o $@
+%.o: %.cpp $(HEADERS)
```
garrettwrong, line 83: note, I had to move this and some other of the more specific patterns to precede the more generic patterns for old systems (like manylinux2010's based on centos6).  I think is it because they are using an ancient gnu make (3.8) which may have slightly different rule evaluation, but I never found the exact documentation about what changed (nor did I look, I've seen this before).

Regardless, this order works for everything I have thrown at it.

README.md
```diff
@@ -4,29 +4,51 @@ A GPU implementation of 2, 3 dimension type 1, 2 non-uniform FFT based on [FINUF
...
+ - Get this code -
  ```git clone https://github.com/flatironinstitute/cufinufft.git```
  - Review the `makefile`. - If you need to customize build settings, create and edit a `make.inc`.  Example:
    - To override the standard CUDA `/usr/local/cuda` location your `make.inc` should contain: ```CUDA_ROOT=/your/path/to/cuda```.
    - Two examples are provided, one for IBM machines (make.inc.power9), and another for the Courant Institute cluster (make.inc.CIMS).
  - Compile - ```make all -j```
  - Run a test code - ``` ./cufinufft2d1_test 2 128 128 10 1e-6```
+ - Many checks can be run using `make check`.
+
+### Library Installation
```
garrettwrong, line 17: I need to update this documentation wrt the expectation of a pypi package for python purposes. That should avoid library conversation except for HPC users.

janden:
Ok so this one is up next? Let me know when it's ready for review.

garrettwrong:
Yes, it is next :).  Either we wait for me to polish the wheels and push them to pypi first, or , we merge this into master after review and worry about the wheels after. (This would just leave few sentences in the README ahead of reality for a few days).

LMK...

I'll undraft either way, but I need to go do some chores this evening.

garrettwrong:
Note that, besides another rebase (and the last 7 commits or so) this is the code in the wheel I sent you yesterday.

Note I did not use the `build-wheels.sh` script to make the wheel (instead I ran the commands by hand). I just started on adopting the script from the PYPA example.

janden:
>  Either we wait for me to polish the wheels and push them to pypi first, or , we merge this into master after review and worry about the wheels after. 

I think we should wait with the wheels a little bit. I'd like to make sure the Python interface aligns with that of Finufft, which has changed a bit over the last month or so.

garrettwrong:
Sure. I saw your list in #38, I'd be happy to cover those. I can do it here or separately afterwards.  I'll just go with the flow if you have time for this. (If you do not, I'll just keep going here and we can get back to it when you're available again).

janden:
I think it'll make most sense to push through with your changes first, then we can go back to #38 and fix them one by one. Should be relatively small. There's also a corresponding list for the Python interface, but since that hasn't completely stabilized in Finufft yet I wanted to wait with that.

README.md
```diff
@@ -3,30 +3,70 @@ A GPU implementation of 2, 3 dimension type 1, 2 non-uniform FFT based on [FINUF
...
    - Two examples are provided, one for IBM machines (make.inc.power9), and another for the Courant Institute cluster (make.inc.CIMS).
  - Compile - ```make all -j```
- - Run a test code - ``` ./cufinufft2d1_test 2 128 128 10 1e-6```
+ - Run test codes - ```make check```
+
+### Library Installation
+
+It is up to the user to decide how exactly to link or otherwise install the libraries produced in `lib`.
+If you plan to use the Python wrapper you will minimally need to extend your `LD_LIBRARY_PATH`,
+such as with `export LD_LIBRARY_PATH=${PWD}/lib:${LD_LIBRARY_PATH}` or a more permanant installation
```
janden, line 23: permanent

README.md
```diff
@@ -3,30 +3,70 @@ A GPU implementation of 2, 3 dimension type 1, 2 non-uniform FFT based on [FINUF
...
+General Python users, or Python software packages which would like to automatically
+depend on cufinufftpy using `setuptools` may use a precompiled binary distribution.
+This totally avoids installing from source and managing libraries for supported systems.
+
+Because binary distributions are specific to both hardware and software,
+we currently only support systems covered by `manylinux2010` that are using
+CUDA 10.1, 10.2, or 11.0-rc with a compatible GPU. This is currently the most
+common arrangement.  If you have such a system, you may run:
+
+`pip install cufinufftpy`
```
janden, line 51: Is there going to be an option here to specify the CUDA version (like `pip install cufinufftpy[cuda102]` or something like that)?

README.md
```diff
@@ -3,30 +3,70 @@ A GPU implementation of 2, 3 dimension type 1, 2 non-uniform FFT based on [FINUF
...
+
+For other cases, the python wrapper should be able to be built from source.
```
janden, line 53: Python

cufinufftpy/cufinufft.py
```diff
@@ -0,0 +1,198 @@
...
+            raise TypeError("cufinufft execute expects {} dtype arguments "
+                            "for this plan. Check plan and arguments.".format(
+                                self.complex_dtype))
+
+        ier = self._exec_plan(c.ptr, fk.ptr, self.plan)
+
+        if ier != 0:
+            raise RuntimeError('Error executing plan.')
+
+    def destroy(self):
```
janden, line 190: Should this just be `__del__`?

setup.py
```diff
@@ -0,0 +1,52 @@
...
+with open(os.path.join('cufinufftpy', 'requirements.txt'), 'r') as fh:
+    requirements = [item.strip() for item in fh.readlines()]
+
+# Sanity check that we can find the CUDA cufinufft libraries before we get too far.
+try:
+    lib = ctypes.cdll.LoadLibrary('libcufinufftc.so')
+except Exception as e:
+    print(lib, 'CUDA shared libraries not found in library path.'
+          '  Please refer to installation documentation at http://github.com/flatironinstitute/cufinufft'
+          ' and ensure CUDA installation is successful first before attempting to install the python wrappers.')
```
janden, line 18: Python (also below)

README.md
garrettwrong: Not right now.  For now, I think it is sufficient to say something like `spam==1.0` is prebuilt for CUDA X `spam==1.23` is prebuilt for CUDA Y. Currently I expect the wheel to work for the release of CUDA that has been dominant for the last year (10), and the one that will be out for the next year or so (11).

If this becomes really popular and there is buy in to support it, I can decouple and make a distribution packaging matrix. This would be like `cufinufftpy==2.0+cuda120` or similar.  That is more work than I can justify right now.

cufinufftpy/cufinufft.py
garrettwrong: Good point, I think I used `__del__` when I was hacking around in ASPIRE. I forgot :). Changed to `__del__`. I removed `plan.destroy()` from the pytests and confirmed they are still running.

In practice we may find that we do not want to wait for python to clean these up , but that is an optimization. That is, I'm not sure how long after the plan instance ref count is zero these will hang around before `__del__` is called.  If it is immediate that's great, otherwise, you can imagine having a loop...  I think it will be pretty obvious if a problem..

garrettwrong:
I think I covered all the suggested changes with the two commits above.

Regarding the Dockerfiles, it was painful but for different reasons :). If we ever want to package gpu-able ASPIRE wheels, it will need the same, so ahead of the game there... We can talk about it another time, but it would actually be totally reasonable to release ASPIRE as a container, particularly if you are targeting corporate (Pharmie companies) or other people that would be running via cloud etc.

README.md
janden: Ok that makes sense. The risk is that users running on clusters (and other machines for which they don't have root access) may not have a choice in their CUDA version. That being said, they can just compile their own for now.

cufinufftpy/cufinufft.py
janden: On the other hand, the user can just call `del plan` in that case, right? This just allows for an automatic (albeit potentially problematic) fallback.

PR #9, opened 2020-05-26
Pass opts by pointer to cufinufft_default_opts

janden:
Instead of by reference. This will hopefully simplify the interface enough to allow for a unified C and C++ interface once we sort out the pointer stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This is a work from Melody Shih's internship at Flatiron Institute, advised by C
  
 ### Interface
 cuFINUFFT API contains 5 stages:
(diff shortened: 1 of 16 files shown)


garrettwrong:
I tried this with #3, but ran into some other things.  Perhaps it is still worth doing, but I think there is a lot more involved in a straight C api than the pointers and extern C'ing the few ctyped methods. I bring it up mainly as crossref.  The pointers are not sufficient to use the C python tools with the code.  Maybe after the other cleanup it is less work.

janden:
> I think there is a lot more involved in a straight C api than the pointers and extern C'ing the few ctyped methods

Of course (there's all the business in utils and the spread options, as you know). What I mean to say is that this should (hopefully) bring us in the right direction.

PR #10, opened 2020-05-26
Unify tests and add help strings

janden:
These should make the test codes a little more consistent with each other and add more extensive help strings.

### DIFF
--- a/test/cufinufft2d1_test.cu
+++ b/test/cufinufft2d1_test.cu
@@ -13,11 +13,16 @@ int main(int argc, char* argv[])
 {
 	int N1, N2, M, N;
 	if (argc<4) {
-		fprintf(stderr,"Usage: cufinufft2d1_test method N1 N2 [M [tol]]\n");
-		fprintf(stderr,"Details --\n");
-		fprintf(stderr,"method 1: nupts driven\n");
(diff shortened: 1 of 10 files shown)


MelodyShih:
Thanks for the docs. They look nice. There are 2 places that might need to be changed:
- The nupts distribution in `spread_3d.cu` is not corrected. I did some tests there and `nupts_distr = 0` is no longer uniformly distributed nupts.  I would suggest delete `case 1, 4` and make `case 3` as `case 1`. 

- The other place is in `spread_2d.cu`: under `case 2`,  there is a if statement when `method==6`, that 4 lines are out-of-dat. 

Since I couldn't push further commit to this pull request, would you mind making the changes? Thanks.

janden:
> There are 2 places that might need to be changed.

Thanks. I will take a look.

> Since I couldn't push further commit to this pull request, would you mind making the changes?

That's strange. I clicked “Allow edits by maintainers” and you are listed as an admin for this repo. What error message is it giving you? If you have the changes on another branch somewhere, I can bring them in.

janden:
@MelodyShih I made the changes you asked for (I think). Can you look through to make sure I got it right?

PR #14, opened 2020-05-27
Put the debug flag before setting the opimization level

### LINKED ISSUE #13: Compiler Optimization flags
I noticed this when I was making the timing change:

```
CXXFLAGS= -DNEED_EXTERN_C  -fPIC -O3 -funroll-loops -march=native -g -std=c++11
```

IIRC -g will override previous optimization levels (the -03 on the left in this case).

Because NVCC has its own opt flags, I don't really think this affected CUDA performance, but in case it would (or another layer of your code),  I'll push a patch momentarily.

garrettwrong:
otherwise I think debug opt level resets to -O0. (Figured while I was in there again :) )

Closes #13

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ NVARCH = -arch=sm_70 \
 	-gencode=arch=compute_75,code=sm_75 \
 	-gencode=arch=compute_75,code=compute_75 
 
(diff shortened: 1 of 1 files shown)


garrettwrong:
Alex was correct, this change is not necessary under gnu. Closing.

PR #15, opened 2020-05-27
Use Thrust and remove the dependency of CUB

MelodyShih:
I am able to find all the corresponding functions in Thrust and used them to replace the calls to functions in CUB. These commits removed the dependency of CUB.

### DIFF
--- a/src/2d/spread2d_wrapper.cu
+++ b/src/2d/spread2d_wrapper.cu
@@ -3,8 +3,10 @@
 #include <iomanip>
 #include <assert.h>
 
-#include <cub/device/device_radix_sort.cuh>
-#include <cub/device/device_scan.cuh>
+#include <thrust/device_ptr.h>
+#include <thrust/scan.h>
+//#include <cub/device/device_radix_sort.cuh>
+//#include <cub/device/device_scan.cuh>
 
 #include <cuComplex.h>
 #include "../cuspreadinterp.h"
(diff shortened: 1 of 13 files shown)


src/2d/spread2d_wrapper.cu
```diff
@@ -65,6 +65,15 @@ int cufinufft_spread2d(int ms, int mt, int nf1, int nf2, CPX* h_fw, int M,
 		2*M*sizeof(FLT)+M*sizeof(CUCPX), milliseconds);
 #endif
 
+	if(d_plan->opts.gpu_method == 1){
```
janden, line 68: So this is not related to CUB/Thrust, right? It's fine to include it in this PR, but would be good to separate these things in the future.

janden:
Thank you for doing this. Should simplify dependencies and the overall code.

garrettwrong:
> Nice, thanks! Code looks lighter. I'll try to give it a closer look and build/run it tomorrow (daughter just got home). Thanks for doing this.

The library builds, and the tests run for me.  I get similar results to master for the 2d and 3d tests.  I have a question about the 3d test, but it is unrelated to this PR.

src/2d/spread2d_wrapper.cu
MelodyShih: You're right. That change is not actually related to this pull request. Ok, I will do that in the future.

PR #20, opened 2020-05-28
ntransfcufftplan → maxbatchsize

### LINKED ISSUE #8: Naming and setting `ntransfcufftplan`
I find this name a little confusing. After digging through the source code, it appears to be the “block size” of the algorithm, that is, the number of inputs to process simultaneously (which, is therefore equal to the number of inputs in the CUFFT plan).

Maybe something like `block_size` would be better? Or `ntransfblock`? We should also discuss good rules of thumb for setting this by default since most users won't want to deal with it. In the tests, we default to 8, but I saw much better performance on my laptop for 16 and 32, after which it seems to saturate.

janden:
Closes #8.


janden:
Pinging @garrettwrong since this changes the API. ABI should be the same, it's just renaming a variable, so your interfaces will hopefully stay the same.

PR #23, opened 2020-05-28
Update default makefile target

### LINKED ISSUE #17: Default target in Makefile
Right now the default target is `spread2d` (since it comes first). Might make more sense to have `all` be the default?

garrettwrong:
Closes #17 .

In #1 I added `clib` to `all`, but forgot to make all the default. This fixes that and adds `.PHONY` for recipe only targets.


garrettwrong:
(note I rebased after adding phony `lib`)

PR #27, opened 2020-05-31
3d1 blockgather

MelodyShih:
These commits fix the error when using spreading method 4 in 3D. The method currently only works for size of the transformation being a multiplication of the size of `obin`. I added some checks so that the codes will return error messages when the condition doesn't satisfy. 

Here is an example output:
image
I set the `obinsize` to (8x8x8) so one can test with size of the transformation being (4Nx4Nx4N), when N being an integer. If the size is not (4Nx4Nx4N), the code will return error.

### DIFF
--- a/contrib/common.cpp
+++ b/contrib/common.cpp
@@ -20,14 +20,19 @@ int setup_spreader_for_nufft(spread_opts &spopts, FLT eps, cufinufft_opts opts)
   return ier;
 } 
 
-void set_nf_type12(BIGINT ms, cufinufft_opts opts, spread_opts spopts, BIGINT *nf)
+void set_nf_type12(BIGINT ms, cufinufft_opts opts, spread_opts spopts, 
+				   BIGINT *nf, BIGINT bs)
 // type 1 & 2 recipe for how to set 1d size of upsampled array, nf, given opts
 // and requested number of Fourier modes ms.
 {
   *nf = (BIGINT)(opts.upsampfac*ms);
   if (*nf<2*spopts.nspread) *nf=2*spopts.nspread; // otherwise spread fails
-  if (*nf<MAX_NF)                                 // otherwise will fail anyway
-    *nf = next235even(*nf);                       // expensive at huge nf
+  if (*nf<MAX_NF){                                // otherwise will fail anyway
+  	if (opts.gpu_method == 4)                     // expensive at huge nf
+    	*nf = next235beven(*nf, bs);
+	else
+        *nf = next235even(*nf);
+  }
 }
 
 void onedim_fseries_kernel(BIGINT nf, FLT *fwkerhalf, spread_opts opts)
--- a/contrib/common.cpp
+++ b/contrib/common.cpp
@@ -28,10 +28,10 @@ void set_nf_type12(BIGINT ms, cufinufft_opts opts, spread_opts spopts,
   *nf = (BIGINT)(opts.upsampfac*ms);
   if (*nf<2*spopts.nspread) *nf=2*spopts.nspread; // otherwise spread fails
   if (*nf<MAX_NF){                                // otherwise will fail anyway
-  	if (opts.gpu_method == 4)                     // expensive at huge nf
-    	*nf = next235beven(*nf, bs);
-	else
-        *nf = next235even(*nf);
+    if (opts.gpu_method == 4)                     // expensive at huge nf
+      *nf = next235beven(*nf, bs);
+    else
+      *nf = next235beven(*nf, 1);
   }
 }
 
--- a/contrib/utils.cpp
+++ b/contrib/utils.cpp
@@ -81,25 +81,6 @@ void arraywidcen(BIGINT n, FLT* a, FLT *w, FLT *c)
   }
 }
 
-BIGINT next235even(BIGINT n)
-// finds even integer not less than n, with prime factors no larger than 5
-// (ie, "smooth"). Adapted from fortran in hellskitchen.  Barnett 2/9/17
-// changed INT64 type 3/28/17. Runtime is around n*1e-11 sec for big n.
-{
-  if (n<=2) return 2;
-  if (n%2 == 1) n+=1;   // even
-  BIGINT nplus = n-2;   // to cancel out the +=2 at start of loop
-  BIGINT numdiv = 2;    // a dummy that is >1
-  while (numdiv>1) {
-    nplus += 2;         // stays even
-    numdiv = nplus;
-    while (numdiv%2 == 0) numdiv /= 2;  // remove all factors of 2,3,5...
-    while (numdiv%3 == 0) numdiv /= 3;
-    while (numdiv%5 == 0) numdiv /= 5;
-  }
-  return nplus;
-}
-
 BIGINT next235beven(BIGINT n, BIGINT b)
 // finds even integer not less than n, with prime factors no larger than 5
 // (ie, "smooth") and is a multiple of b (b is a number that the only prime 
--- a/contrib/utils.cpp
+++ b/contrib/utils.cpp
@@ -100,6 +100,27 @@ BIGINT next235even(BIGINT n)
   return nplus;
 }
 
+BIGINT next235beven(BIGINT n, BIGINT b)
+// finds even integer not less than n, with prime factors no larger than 5
+// (ie, "smooth") and is a multiple of b (b is a number that the only prime 
+// factors are 2,3,5). Adapted from fortran in hellskitchen. Barnett 2/9/17
+// changed INT64 type 3/28/17. Runtime is around n*1e-11 sec for big n.
+// added condition about b Melody 05/31/20
+{
+  if (n<=2) return 2;
+  if (n%2 == 1) n+=1;   // even
+  BIGINT nplus = n-2;   // to cancel out the +=2 at start of loop
+  BIGINT numdiv = 2;    // a dummy that is >1
+  while ((numdiv>1) || (nplus%b != 0)) {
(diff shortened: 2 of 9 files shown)


ahbarnett:
Sorry to say something silly, since I haven't followed all the gpu details. But:
With the cpu version, the user requests *arbitrary* mode numbers (N1,...) and the library makes an internal decision that is convenient for spreading/fft, etc.
This is done for each dimension by
set_nf_type12
which writes into nf1, .. etc.  It picks nf big enough for spread, and a multiple of 2,3,5.
It looks like you do the same here in
https://github.com/flatironinstitute/cufinufft/blob/e3f5b924bb8c3ee21c5b5b1d72834677b07bf3d0/src/cufinufft.cu#L59

So, shouldn't you just edit set_nf_type12 to also return each nf a multiple of obin?
Best, Alex

MelodyShih:
Hi Alex, that is a great point. You're totally right about that I can just edit the set_nf_type12 function and set nf to a number that is multiple of obinsize. Thanks for pointing this out!

MelodyShih:
I modified the code (commit b97e972) according to Alex's comment. Now block gather method (gpu_method=4) also works for arbitrary size of transformation. :)

image

src/3d/spread3d_wrapper.cu
```diff
@@ -20,7 +20,18 @@ int CalcGlobalIdx(int xidx, int yidx, int zidx, int onx, int ony, int onz,
 	oix = xidx/bnx;
 	oiy = yidx/bny;
 	oiz = zidx/bnz;
-	return   (oix+oiy*onx + oiz*onx*ony)*(bnx*bny*bnz) +
+#if 0
```
garrettwrong, line 23: Do you have a DEBUG sort of flag you are using elsewhere? Some people dislike leaving debug prints around, but generally I don't mind if it is not disruptive to program because it helps understand what the developer was thinking or struggling with.  

However, if we leave it, its nice that it is pretty standard through out a code, and maybe gets tested once and a while, which again, is easier if there is a flag(s) to flip.

Do you think this is something that is useful to keep around for the next developer (which might be you :) )?

contrib/utils.cpp
```diff
@@ -100,6 +100,27 @@ BIGINT next235even(BIGINT n)
   return nplus;
 }
 
+BIGINT next235beven(BIGINT n, BIGINT b)
```
janden, line 103: Can `next235even` be written to call this function?

contrib/common.cpp
```diff
@@ -20,14 +20,19 @@ int setup_spreader_for_nufft(spread_opts &spopts, FLT eps, cufinufft_opts opts)
...
 // type 1 & 2 recipe for how to set 1d size of upsampled array, nf, given opts
 // and requested number of Fourier modes ms.
 {
   *nf = (BIGINT)(opts.upsampfac*ms);
   if (*nf<2*spopts.nspread) *nf=2*spopts.nspread; // otherwise spread fails
-  if (*nf<MAX_NF)                                 // otherwise will fail anyway
-    *nf = next235even(*nf);                       // expensive at huge nf
+  if (*nf<MAX_NF){                                // otherwise will fail anyway
+  	if (opts.gpu_method == 4)                     // expensive at huge nf
+    	*nf = next235beven(*nf, bs);
```
janden, line 32: Fix indentation.

src/3d/spread3d_wrapper.cu
```diff
@@ -452,7 +463,19 @@ int cuspread3d_blockgather_prop(int nf1, int nf2, int nf3, int M,
 	int o_bin_size_x = d_plan->opts.gpu_obinsizex;
 	int o_bin_size_y = d_plan->opts.gpu_obinsizey;
 	int o_bin_size_z = d_plan->opts.gpu_obinsizez;
+	
 	int numobins[3];
+	if (nf1 % o_bin_size_x != 0 ||
```
janden, line 468: Is this check (and the one below) still necessary given the fact that we make sure to have the correct `nf*`?

src/3d/spread3d_wrapper.cu
```diff
@@ -644,6 +684,7 @@ int cuspread3d_blockgather_prop(int nf1, int nf2, int nf3, int M,
 	checkCudaErrors(cudaMemcpy(&totalNUpts,&d_binstartpts[n],
 		sizeof(int),cudaMemcpyDeviceToHost));
 	checkCudaErrors(cudaMalloc(&d_idxnupts,totalNUpts*sizeof(int)));
+	checkCudaErrors(cudaMemset(d_idxnupts,-1,totalNUpts*sizeof(int)));
```
janden, line 687: Why this change?

test/cufinufft3d1_test.cu
```diff
@@ -100,10 +100,29 @@ int main(int argc, char* argv[])
...
-	dplan.opts.gpu_binsizex = 16;
-	dplan.opts.gpu_binsizey = 16;
-	dplan.opts.gpu_binsizez = 2;
-	dplan.opts.gpu_maxsubprobsize = 4096;
+	dplan.opts.gpu_kerevalmeth=1;
+	switch(method){
+		case 1:
+		case 2:
+			{
+				dplan.opts.gpu_binsizex = 8;
```
janden, line 108: Couldn't these things be folded into the default options? Is it even desirable for users to adjust these? I doubt anyone besides @MelodyShih is going to fiddle with these.

src/3d/spread3d_wrapper.cu
MelodyShih: Hi Garrett, sorry for the late reply. Yes, the debug codes should at least be standard through out the the code. I'll remove statements like `#if 0` in another pull request. 

I think for now it would be nice to keep the codes inside `#ifdef DEBUG`. They are helpful for me to see what went wrong through out the computation and might also helpful for other potential developers. But if you have better idea about how to manage the debug statements, please let me know, thanks!

contrib/utils.cpp
MelodyShih: Yes, they can (see 5f75d96).

contrib/common.cpp
MelodyShih: (See 5f75d96)

src/3d/spread3d_wrapper.cu
MelodyShih: I leave it there because there is function that only does the step of spreading (`cufinufft_spread3d`) and there, `nf*` is an input from users.

src/3d/spread3d_wrapper.cu
MelodyShih: That is actually for debugging so that I can make sure all entries that should be set are set. I make it a debug statement in commit 2c6909b.

test/cufinufft3d1_test.cu
MelodyShih: The default method for 3d type 1 is the sub problem method (gpu_method=2) and so the default value for `gpu_binsize*` are set for that method. Since `cufinufft_default_opts` is called before the method is set, I am not sure how to do it properly. Please let me know if you have and suggestions, thanks!

contrib/utils.cpp
janden: Great. Thank you.

src/3d/spread3d_wrapper.cu
janden: Ok sounds good.

src/3d/spread3d_wrapper.cu
janden: Ok.

test/cufinufft3d1_test.cu
janden: I think these things are so far in the weeds that we may want to hardcode them (with good values). I don't think anyone will start changing these. That way we can also simplify the opts structure.

If we do want to keep them, an option would be to have the “default” bin sizes (as set by `cufinufft_default_opts`) set to zero, (or negative one) which means that the hardcoded values are used.

MelodyShih: I think I would still like to keep them as options so that users can overwrite them once they get more familiar with the methods. Setting the options to -1 seem to be a good idea to resolve the problem! Thanks for the suggestion. I did it in commit 2a2e51d. Now, there is a check in the plan stage that if the options are -1, then some hard coded values are used (see function `set_binsize`).

test/cufinufft3d1_test.cu
```diff
@@ -101,6 +101,8 @@ int main(int argc, char* argv[])
 	ier=cufinufft_default_opts(type, dim, &dplan.opts);
 	dplan.opts.gpu_method=method;
 	dplan.opts.gpu_kerevalmeth=1;
+
+	/* if not set, some hard coded values for the bin size are used */
```
janden, line 105: Won't the values go away for method 1 and 2 as well?

MelodyShih: Yes, I am thinking of having them there as an example to show that users can set the options to values they'd like to try.

MelodyShih: but if you think this is not necessary and somehow confusing, I can remove them too.

janden: I think it might be confusing. Most readers (including myself) will read this as “you have to do this” as opposed to “you can do this”. Again, I expect a very small portion of users will actually want to fiddle with this.

MelodyShih: I see. Thanks for the feedback!  I removed them in 3988f01.

janden: Great!

Issue #19, opened 2020-05-28
TIME flag not consistently treated in tests


janden:
Even with the flag turned off, these codes output certain timings. Since these are only there for testing purposes, it might make sens for them to ignore the TIME flag altogether.

garrettwrong:
I noticed this when I ran the example test, but it looked like the methods ASPIRE was calling were silent, so I assumed it was just intentionally there for the examples (which seemed fine).  Did you get output somewhere you thought it definitely shouldn't be or is this just for consistency?

I ask because if we're leaking, it can blow up logs in ASPIRE pretty quick 😅 .

janden:
> it looked like the methods ASPIRE was calling were silent

Yes that seems good. The timing scripts are using the TIME flag to determine whether to output their own timing (wrapping the library calls). This is just for consistency. When it's disabled, some parts are timed while others aren't. It would be simpler for them to always output timings, since they're essentially timing code.s

MelodyShih:
The places TIME flag are used in the test codes are around a call of cudaMalloc(&a,1), which is now redundant. I was using it to "warm up" GPU but since we exclude the time allocating the GPU memory, it is no longer needed. I will delete those codes and create a pull request.

PR #29, opened 2020-06-07
removing TIME flags in the test codes. 

MelodyShih:
This is related to ISSUE #19 . 

Places that still use TIME flags are:
- Around calls of cudaMalloc(&a,1). 
The call of this function is no longer needed since we are now excluding the timing of allocating the gpu memory. I will delete the call and also the TIME flags for this part. But since for spread and interpolation only tests, we still allocate the gpu memory during the call of `cufinufft_spread*d` and `cufinufft_interp*d`. I leave the warm up code for those tests but delete the TIME flags around it. 
- Inside cufinufft2d2_test.cu. To be consistent to other tests, we still time the code even if the user doesn't set the TIME flags.

### DIFF
--- a/test/cufinufft2d1_test.cu
+++ b/test/cufinufft2d1_test.cu
@@ -79,17 +79,6 @@ int main(int argc, char* argv[])
 	cudaEventCreate(&start);
 	cudaEventCreate(&stop);
 
-	/*warm up gpu*/
-	cudaEventRecord(start);
-	char *a;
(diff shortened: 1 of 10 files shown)


janden:
Looks good. So now all the tests will output timings regardless of whether TIME is specified?

MelodyShih:
Yes, that is the case. (I found that the nupts case doesn't match the argument descriptions --- 185691f)

PR #30, opened 2020-06-12
using only DEBUG flag for debug codes and remove unused codes

MelodyShih:
These commits contains following changes:

- In the test codes, there are some not-used codes for outputting the results, which is outdated. I removed them and only remains those for spread/interp only tests. (1addcaa)
- For 2d1many and 2d2many tests, randomly picked up one data and test the accuracy. (1addcaa)
- Some kernels and functions are codes that I wrote before for trying other methods. They are no longer used now so I removed them (667bb98, d35067a)
- Removed some codes for debugging and use only DEBUG flag for debug codes (a6b0c28)


test/cufinufft2d1many_test.cu
```diff
@@ -149,30 +149,16 @@ int main(int argc, char* argv[])
...
-		for (int i=0; i<nf1; i++){
-			//                if( i % opts.gpu_binsizex == 0 && i!=0)
-			//                        printf(" |");
-			printf(" (%2.3g,%2.3g)",fw[i+j*nf1].real(),fw[i+j*nf1].imag() );
-		}
-		cout<<endl;
-	}
-#endif	
+
+	int i = (int) (0.13*ntransf); // // choose some data to check
```
garrettwrong, line 153: If I read correctly, this implies we need a minimum ntransf to run the tests in a way that won't test only the first matrix in the batch. It might be worth bumping it to at least the second for any case where ntransf>1.

I am thinking the zero'th matrix is the most likely to work if someone changes the code incorrectly later.

ahbarnett: Suggest change to int i = ntransf-1;
This always tests the last one.

MelodyShih: Thanks for the suggestion. I will change it to test the accuracy of the last data.

MelodyShih:
Have to admit there’s always hesitation to get rid of things but I’m happy with the clean-up results:)

janden:
Well, it's still in the history so you can always go back…!

PR #31, opened 2020-06-14
Add a simple collection of run checks to makefile

garrettwrong:
I have been using something like this from my own text file, adjusting it here and there.

@MelodyShih you probably have the best insight into which tests are meaningful (sizes) and would cover (exercise) the most out of your code. I think some of these sizes I wrote down relate to ASPIRE, though the NU points might not. 

Please change it up as you see fit :) Don't be afraid to add run tests, even if they duplicate a little, to exercise your code. These really don't take too long, even on my older cards.

Just having the make check will probably save more time than extra test lines take :).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,59 @@ clean:
 	rm -rf lib
 	rm -rf lib-static
 
+check2D: all
+	@echo Running 2-D cases
+	bin/cufinufft2d1_test 1 256 256
+	bin/cufinufft2d1_test 2 512 512
+	bin/cufinufft2d2_test 1 256 256
+	bin/cufinufft2d2_test 2 512 512
+	@echo Running 2-D High Density cases
+	bin/cufinufft2d1_test 1 64 64 8192
+	bin/cufinufft2d1_test 2 64 64 8192
+	bin/cufinufft2d2_test 1 64 64 8192
+	bin/cufinufft2d2_test 2 64 64 8192
+	@echo Running 2-D Low Density cases
+	bin/cufinufft2d1_test 1 64 64 1024
+	bin/cufinufft2d1_test 2 64 64 1024
+	bin/cufinufft2d2_test 1 64 64 1024
+	bin/cufinufft2d2_test 2 64 64 1024
+	@echo Running 2-D-Many cases
+	bin/cufinufft2d1many_test 1 64 64 128 1e-3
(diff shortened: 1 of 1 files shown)


ahbarnett:
looks good. Might want to add tol = 1e-3 vs tol = 1e-6, and a high vs low density (M >> N1.N2 vs M << N1.N2, for example, in 2d). But that's more like performance tests.

I have found a "bad inputs" test very useful for debugging and finding segfaults.
See test/dumbinputs in FINUFFT.

garrettwrong:
Cool, thanks. I added some runs like you suggested.

Making an issue for the faulty input sort of checks; that is a good idea.

Because there are now quite a few lines, I broke into 2D and 3D with an if PREC statement adding 3D when PREC is SINGLE.

Once this is in, if you have a gpu-able CI platform for FI codes, they could run this as a sanity check.  It isn't fancy, but once automated it is more than a lot of codes have. It will also facilitate if you don't have a GPU handy, just push the code up and CI will build it for you...

```
# make all and run checks in doubles (2D)
make all check
# make all (force recompiles) in singles and run checks in singles (2D, 3D)
PREC=single make -B check
```

MelodyShih:
Thanks for setting this up. They are helpful. I added more tests to the makefile and even found some bugs from the 3d codes... You might meet some problems running the added tests because of that. The error is fixed in #33. The change should make all the tests code passed. Please let me know if that is not the case on your machine.

garrettwrong:
Can confirm these additional catch those typos now fixed in the other PR.

```
bin/cufinufft3d1_test 4 1e2 2e2 3e2 1e4
[time  ] cufinufft plan:		 0.105 s
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  scan failed to synchronize: cudaErrorIllegalAddress: an illegal memory access was encountered
make[1]: *** [Makefile:2
```

Locally I merged with #33 and #30 and everything seems good together :). 

I suspect you will find automating this really satisfying...

ahbarnett:
The joy of testing rectangular not square cases :)

janden:
> Once this is in, if you have a gpu-able CI platform for FI codes, they could run this as a sanity check.

[There is](https://jenkins.flatironinstitute.org). I can talk to Dylan about setting this up. Should be very useful.

PR #35, opened 2020-06-15
return ier error code so upstream code knows there was a problem

### LINKED ISSUE #34: Propagation of shared memory error
Making the dual type python wrapper, I ran into a case where the code wasn't handling an error correctly.  If I do something I shouldn't, like call a 3d method with doubles I get the shared memory allocation errors printed,

```
error: not enough shared memory (92160)
error: cnufftspread3d_gpu_subprob
error: cuspread3d, method(2)
```

but from `spread3d_wrapper.cu` we begin returning 0, so calling code doesn't raise as an error.  Probably we should return a non zero code.  I took a quick look around and picked up a few more, but I might have missed some.  Its minor change, will push momentarily.

garrettwrong:
Closes #34 .

From the python side, taking this code allows us to raise something like below instead of continuing (and failing later on bad data).
```
E           RuntimeError: Error executing plan.
```


ahbarnett:
I have just been updating error/warning reporting in finufft (guruInterface_v2, which will go to master in a week or so) too.

In C++, I keep the f77-style "return ier" status reporting, but I allowed that some nonzero ier codes can be warnings only (for now, just when tol is too small, which used to be a fatal error, now raises warning ier=1 but does the transform, which can be trusted). Others cause exit.

In matlab wrapper I added errhandler.m to convert this to matlab-style raising an error or warning, but removed any ier returned as a numerical output argument. This is matlab style.

I haven't looked at python error-reporting style yet but it would be good to rationalize this with Libin @lu1and10 in finufft.  I believe raising an error rather than python interface returning an ier argument is also the way to go.

garrettwrong:
The python bindings will raise on non zero ier. They don't differentiate for any warning level items, though they certainly could if logic and an enum or something is added.

We don't pass the `ier` outside the cufinufft python code (ie do not pass through interface).  cufinufft is just expected to raise a python exception as you describe.  This just patches a couple cases where that `ier` wasn't making it all the way out of c++.

garrettwrong:
They print unless we want to explicitly control them. I actually think they are helpful. I wouldn't do anything about it. I don't think this is particularly bad practice or anything like that. It's not like a memory leak or anything.

janden:
> I actually think they are helpful. I wouldn't do anything about it.

Ok sounds good. Just checking.

PR #39, opened 2020-07-20
Bring closer to latest guru v2

### LINKED ISSUE #38: Harmonize C interface with that of Finufft
The current interface was modeled on an early version of the Finufft plan interface, which has changed in the meantime.

- [x] Options passed to `cufinufft_makeplan`. Currently, the user has to set these manually as part of `plan`. We should follow Finufft in allowing the user to provide the options to `cufinufft_makeplan`.
- [x] Plan is a pointer to a struct as opposed to a struct. This should simplify the ctypes interface, among other things.
- [x] Naming. Really only `cufinufft_setNUpts` that should become `cufinufft_setpts`.
- [x] Merge C++ and C interfaces. Currently, we have `cufinufft` and `cufinufftc`. Ideally, we'd only have one. I know there have been issues with this in the past, but if we pass the options to `makeplan` and make the plan a pointer, a lot of the C++ dependency can be stashed away from the API.

I'm sure there are more things, but these are the first that come to mind.

garrettwrong:
Closes #38 

This PR is currently on top #6, and is just from https://github.com/flatironinstitute/cufinufft/commit/113c00c44860bd88e2a0b68d2d5106c9abfa8222 forward.


Makefile
```diff
@@ -247,7 +231,6 @@ check3D_64: all
 	bin/cufinufft3d1_test 4 15 15 15 2048 1e-3
 	bin/cufinufft3d2_test 1 16 16 16 4096 1e-3
 	bin/cufinufft3d1_test 1 128 128 128
-	bin/cufinufft3d1_test 2 16 16 16
```
janden, line 250: Why gone?

cufinufftpy/_cufinufft.py
```diff
@@ -112,135 +112,54 @@ def _get_SpeadOptsFields(dtype):
     return fields
 
 
-class SpreadOpts(ctypes.Structure):
-    pass
-
-
-SpreadOpts._fields_ = _get_SpeadOptsFields(np.float64)
```
janden, line 119: Remove `_get_SpreadOptsFields` since it is no longer used (AFAICT).

cufinufftpy/_cufinufft.py
```diff
@@ -112,135 +112,54 @@ def _get_SpeadOptsFields(dtype):
...
-
-CufinufftPlan._fields_ = _get_CufinufftPlan(np.float64)
-
-
-class CufinufftPlanf(ctypes.Structure):
-    pass
-
-
-CufinufftPlanf._fields_ = _get_CufinufftPlan(np.float32)
+CufinufftPlan = c_void_p
```
janden, line 115: :smile:

test/cufinufft2d1_test.cu
```diff
@@ -141,7 +151,7 @@ int main(int argc, char* argv[])
 		cudaMemcpyDeviceToHost));
 
 	printf("[Method %d] %ld NU pts to #%d U pts in %.3g s (\t%.3g NU pts/s)\n",
-			dplan.opts.gpu_method,M,N1*N2,totaltime/1000,M/totaltime*1000);
+			dplan->opts.gpu_method,M,N1*N2,totaltime/1000,M/totaltime*1000);
```
janden, line 154: `opts.gpu_method` should suffice here, right? Don't want to be dereferencing `dplan` unless we absolutely have to.

Makefile
garrettwrong: 3D method 2 has the smem limit.  There is a comment a few lines above to that effect.

It was actually erroneous before, but it wasn't picked up until I started returning `ier`...

cufinufftpy/_cufinufft.py
garrettwrong: Yes nice catch.  I should have ran a linter on the python code. I just did and it caught a few similar things which I've fixed. Hope you don't mind (they're all very minor changes).

test/cufinufft2d1_test.cu
garrettwrong: Correct, good point.

cufinufftpy/_cufinufft.py
garrettwrong: that plan came together nicely

Makefile
janden: Ah right. Forgot about that.

cufinufftpy/_cufinufft.py
janden: :+1:

cufinufftpy/_cufinufft.py
janden: :grimacing:

test/cufinufft2d1_test.cu
janden: There are a few more occurrences of this, otherwise we're good here.

garrettwrong: I fixed the two remaining in tests.  There are many refs internally in `src/` but the bunch I looked at appear to be cases where the function is only passed a `CUFINUFFT_PLAN` (no other opts struct).

janden: That's totally fine. Just the public API stuff that we should be careful with here.

PR #40, opened 2020-07-23
Rename execute

garrettwrong:
Will undraft after when #39 is in.  Will need rebase the PR. (this will just be https://github.com/flatironinstitute/cufinufft/pull/40/commits/8b100a682793d8783f3c5287949af8a79a91defd )

This just makes the method name consistent with a change in FINUFFT.


garrettwrong:
Sorry for all the action @MelodyShih , I think we're calming down soon :)

MelodyShih:
@garrettwrong , not at all a problem. :) Thanks for all the work!

PR #41, opened 2020-07-24
Add NERSC and OLCF site-sepcifics to make file

JBlaschke:
I am working on deploying `cufinufft` on the emerging ECP systems -- using Summit and Cori as Pre-ECP testbeds. NERSC's Perlmutter to follow (once I get a chance to play with it).

1. On OLCF Summit: `make -j` works just fine
2. On NERSC's Cori system: `make -j site=nersc_cori` (using Cray compiler wrappers -- this _will_ get MPI all wrong on Cori GPU)
3. On NERSC's Cori GPU system `make -j site=nersc_cgpu` (not using Cray compiler wrappers -- for some reason esslurm does not play nicely with the compiler wrappers)


garrettwrong:
Neat, that is exciting.  Everyone has been working hard recently to push this towards being more portable and packaged up :). Perfect timing.

Makefile
```diff
@@ -24,25 +31,61 @@ NVCCFLAGS= -std=c++14 -ccbin=$(CXX) -O3 $(NVARCH) -Wno-deprecated-gpu-targets \
...
 LIBS=-lm -lcudart -lstdc++ -lnvToolsExt -lcufft -lcuda -l$(FFTWNAME) -l$(FFTWNAME)f
+ifdef FFTW_DIR
+    $(info detected a FFTW_DIR variable -- setting FFTW library directory)
+    LIBS += -L$(FFTW_DIR)
+endif
+
+# NVCC-specific includes
+NVCC_INC=
+ifdef FFTW_INC
+    NVCC_INC += -I$(FFTW_INC)
```
garrettwrong, line 62: I think we may not be using this variable?

Makefile
```diff
@@ -24,25 +31,61 @@ NVCCFLAGS= -std=c++14 -ccbin=$(CXX) -O3 $(NVARCH) -Wno-deprecated-gpu-targets \
...
+endif
+
 
 
 #############################################################
 # Allow the user to override any variable above this point. #
--include make.inc
+uname_p := $(shell uname -p)
+ifeq ($(uname_p), ppc64le)
+    -include make.inc.power9
```
garrettwrong, line 81: If a power9 user has a `make.inc` (I would if I was on Traverse at Princeton for example), this may override it.

We might want to always take the `make.inc` if it is available. It doesn't ship with our code, so a user would have to intentionally create or symlink it...

With that said, taking the best default otherwise makes sense to me, in the case `make.inc` dne.  

@ahbarnett , do you have a preference here so we stay inline with FINUFFT? Thanks

JBlaschke:
> Neat, that is exciting. Everyone has been working hard recently to push this towards being more portable and packaged up :). Perfect timing.

Glad to be of help -- I'm trying to follow what I think is the overall design of cufinufft, so let me know if I should change something because it doesn't mesh with where you want to go (with the project overall)

Makefile
JBlaschke: It's set by the `cray-fftw` module on cori.

I can be persuaded to move that into the site-specific `.inc`s

Makefile
JBlaschke: I see where you're going with this: `make.inc` => local user settings. I'm happy with this approach. The design here would then be:

1. `make.inc` -> local user settings
2. `make.inc.arch` -> architecture-specific settings
3. `site/**` -> site (like compute center, etc) specific settings

Makefile
garrettwrong: Right, but I think maybe you intended to use `NVCC_INC` somewhere later in the makefile?

Makefile
garrettwrong: Yes that would be okay by me, but we would want to stay consistent with our related package FINUFFT, so the notion of a `site` directory for make include files should get run by both projects (I tagged the projects' lead for you).  It may be suggested to keep things "flat" for simplicity.  I'm just helping out with the project recently; I'm sure FI team will give you some feedback here (they are eastern time so maybe tomorrow.)

Makefile
JBlaschke: Ah yes -- that got eaten by a merge -- and I don't think we need `NVCC_INC` anymore! I'll Delete it.

Makefile
JBlaschke: Thanks. I'll keep an eye out for what they stay!

ahbarnett: Hello Johannes & Garrett,

I am fine you adding a site directory of more make.inc.* files, as long as you document it.
In our CI for FINUFFT we include eg
`cp make.inc.macosx_clang make.inc`
as part of the docker setup.
So the set of make.inc.* will become more reliable and system-specific.

I don't think the main makefile should have the OS-specific checks though - will become too messy. Maybe a separate script or makefile which sets up make.inc for the system/site is best? I'm not rigid about this though.

Makefile
JBlaschke: Done @garrettwrong -- `NVCC_INC` is gone

Makefile
JBlaschke: Hey @ahbarnett 

I've gone and removed the architecture check for power9 -- technically `uname_p := $(shell uname -p)` checks for the cpu architecture, so I wouldn't call it OS-specific checks (you're already running make 😉 ). But I know what you mean. 

Have a look at the beginning of `Makefile` now -- I've implemented a compromise solution, where the user can specify a site and machine architecture using eg `make site=<computer> target=<archtiecture>`. The sites can also override architectures.

In keeping with the design, I've also moved the targets into a dedicated `targets` directory -- and `make.inc` can be used to overwrite variables locally (as before).

Let me know what you think.

Makefile
```diff
@@ -1,38 +1,85 @@
...
+    INC += -I$(FFTW_INC)
+endif
 
 FFTWNAME=fftw3
 
-LIBS=-lm -lcudart -lstdc++ -lnvToolsExt -lcufft -lcuda -l$(FFTWNAME) -l$(FFTWNAME)f
+# Common libs
+LIBS=-lm -lcudart -lstdc++ -lnvToolsExt -lcufft -lcuda -l$(FFTWNAME)
+
+# Figure out the specifics of the FFTW install
```
ahbarnett, line 63: Why can't we just require the FFTW3 repo, which includes single-prec fftw3f libs? Are there environments with libfftw3 but not libfftw3f ?

Makefile
```diff
@@ -1,38 +1,85 @@
-CC=gcc
-CXX=g++
-NVCC=nvcc
+# Load site-specific setting -- controlled using environment variable `site`:
```
ahbarnett, line 1: I am fine with these new make options. It's clean,and it we're helping users with various setups, great. Don't forget to document these make options in the docs and the makefile. Note in the FINUFFT makefile I did it the other way round, ie set defaults via :=  and then -includes the make.inc.  All good.

Makefile
JBlaschke: The `fftw` package on Cori GPU doe not include the single-prec `libfftwf*` libraries. Since we're not explicitly dependent on the single-prec libraries, and since it's possible to build fftw without them why force to user to include them?

Makefile
JBlaschke: Where would you like me to document this? The README (in lieu of an actual docs -- like the one for NUFFT?)

Makefile
garrettwrong: Hey guys, their inclusion here is just oversight on my part.  This code makes use of `cuFFT`...  FINUFFT makes use of `fftw`.  I think the inclusion here was just oversight in the process of bring them closer together (I was altering makefiles for both). There was a time where there was more dependence on internal structures, but this is not the case now.

I don't think we need fftw shared library at all for cufinufft. This was a good catch. I will take care of removing it in another PR.  For now I think you may proceed with this to facilitate the rest of these changes.  I expect the extra lines would be removed... OR

If you want to wait I can put a PR removing FFTW tomorrow, plus review the time. (Not sure how rushed you are to close this out).

Makefile
garrettwrong: I think the confusion is because cufinufft's documentation is not as comprehensive as FINUFFT.  I took a pass at docs for the Python code (sphinx), but we just have the README for the top level code here. I can imagine one day the documentation being comprehensive for the top level (can just include the python docs then) but that is out of scope for this PR.

It might be reasonable to add an `Advanced Makefile Usage` or something similar as a level 2/3 heading in the README, depending how much detail it requires to be clear for your users.

JBlaschke: Done -- I've put a wee doc into the README

Makefile
garrettwrong: Once #46 is merged, and brought in here, the line/changes for this conversation should go away.

Makefile
garrettwrong: Looks very serviceable, thank you for doing that.

@MelodyShih , note this will change location of the make.inc file I think you used.. so you're not surprised.

MelodyShih: Got it. Thanks for letting me know.

Makefile
JBlaschke: Gotcha -- I'll wait for #46

garrettwrong: FFTW should be removed master now. Hope that helps. Thanks

JBlaschke: Thanks, I've merged the changes from #46 into this PR

JBlaschke:
Hey @garrettwrong @MelodyShih @ahbarnett -- can we get this merged with master at some point?

ahbarnett (approved):
Thanks for the docs - very clear

PR #42, opened 2020-07-24
Misc Patches

garrettwrong:
When integrating the latest in with ASPIRE I noticed a few problems.

My type checking was not correct for 2D. Also we appeared to try and free some junk if `set_pts` was not called first, so I zero'd out the struct which seems to be satisfactory for calming that corner case.

Currently these patches return the UT on the ASPIRE side to working.  ~I'm still chewing through one of our demo codes.  Hopefully this is all I find.~


garrettwrong:
Thanks @MelodyShih , I might add one or two more minor commits. Last time I spoke with Joakim he mentioned about changing the name which is mostly here, but I missed at least one place, I'll push later.   I'm trying to get the python distribution wheels completed and pushed the the PyPI Test index today. I will add a readme file describing how I did that.

I will re-request then. Sorry I thought I could sneak it all in before either of you did your review, I should have re-drafted it.

garrettwrong:
These changes are minor, and only touch readme and the python bindings. I am going to merge them in so we can proceed with PyPI. It is low risk.

PR #46, opened 2020-08-11
Remove dead FFTW code inherited from FINUFFT

### LINKED ISSUE #45: Vestigial Dependence on FFTW
A user had a case where they were have conflicts integrating an FFTW install at a HPC with cufinufft.

FFTW dependence was brought over from FINUFFT in the process of trying to use their headers in contrib. To the best of my knowledge cufinufft uses cuFFT exclusively. Besides the vestigial lines from the parent repo, we don't use FFTW here.

So we need to decide which is more important. Travelling light (removing the unused dependence), or having the contrib files from another project be as similar/identical as possible. Since the files already required changing, I might suggest trimming the dependence.

I will push a PR shortly that removes it for consideration. Its a very small change.

garrettwrong:
Closes #45 , there are a few sentences describing the situation there and in #41 .

I'm not attached to this change either way, just providing the option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,7 @@ INC=-I$(CUDA_ROOT)/include \
 	-Icontrib/cuda_samples
 NVCC_LIBS_PATH=-L$(CUDA_ROOT)/lib64
 
-FFTWNAME=fftw3
-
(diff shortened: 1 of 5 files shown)


JBlaschke:
This is only related in the sense that you've made me think of cuFFT vs a more general FFT backend: I assume that a HIP version is not in the pipeline?

PR #47, opened 2020-08-31
Fix tof cleanup

garrettwrong:
This code tries to ensure that cleanup only occurs ~once~ in the correct order.  The initial `__del__` method wasn't always working correctly atexit from top of file scope (but was working fine in functions etc).

### DIFF
--- a/python/cufinufft/cufinufft.py
+++ b/python/cufinufft/cufinufft.py
@@ -5,6 +5,7 @@ the cufinufft CUDA libraries.
 """
 
 import atexit
+import sys
 
 import numpy as np
 
@@ -22,6 +23,12 @@ from cufinufft._cufinufft import _exec_plan
 from cufinufft._cufinufft import _exec_planf
 from cufinufft._cufinufft import _destroy_plan
 from cufinufft._cufinufft import _destroy_planf
+
+
+# If we are shutting down python, we don't need to run __del__
+#   This will avoid any shutdown gc ordering problems.
+exiting = False
+atexit.register(setattr, sys.modules[__name__], 'exiting', True)
 
 
 class cufinufft:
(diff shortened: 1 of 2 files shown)


janden (changes requested):
So if I understand correctly, this should prevent double deletes? How does using `finalize` achieve that, exactly?

Also, the docs indicate that the finalizer function *should not* be a bound method of the object (“It is important to ensure that func, args and kwargs do not own any references to obj, either directly or indirectly, since otherwise obj will never be garbage collected. In particular, func should not be a bound method of obj.”) So doesn't that cause problems here?

garrettwrong:
You might be right this isn't the best tool for the job. I didn't find any problems testing this, but I didn't make a degenerate test case to check for accumulating these over a long running program either.

finalize gets called exactly once, under all of the circumstances we care about...

I was going to add some logic to handle `__del__` better, but it seemed like this did something similar and was being suggested as standard. I didn't catch that note in the documentation you quoted (after atexit section), but you are correct that doesn't sound right.

I'll convert to draft and dig in a little deeper.  I think the original issue is that on exit, some references are cleaned up in the wrong order.  When we used `destroy()` if a user forgot to call destroy at exit, it was no big deal, but using `__del__` I think we'll need to manage some references one way or another so that destroy is still called correctly. (I hope that makes sense).

janden:
Got it. I agree that `finalize` seems to be the way to go.

garrettwrong:
I can probably figure something simpler out, I was being lazy and you caught me fair and square :D.

garrettwrong:
I think this is simpler, and clearer what should happen.

I tested from some ASPIRE variant I have and seemed fine. It's late, so I'll review with fresh eyes and undraft tomorrow morning.

janden (approved):

PR #57, opened 2020-09-03
Fix error message in setup.py

janden:
When libraries are not found, we can't use lib (since it's not defined).


garrettwrong (approved):

garrettwrong:
Yes this was a vestigial line from when there were two libraries, thanks for fixing it.

Issue #65, opened 2020-09-22
status and precision of test/{spread,interp}{2,3}d ?  bin/interp2d 2 0 ... fails.


ahbarnett:
these testers are compiled but not tested in the `make check`.
Are they supported, or should they go into some `devel` dir instead?
Are they double-prec only?  (not so useful, since most GPU will be sinlge-prec).

I ask because
`bin/interp2d 2 0 1e3 1e3 1e7 1e-3`
fails for me with, eg:
```
(base) ahb@ccmlin001 /mnt/home/ahb/numerics/cufinufft> bin/interp2d 2 0 1e3 1e3 1e7
[time  ] (warm up) First cudamalloc call 2.429e-04 s

CUDA error at src/2d/interp2d_wrapper.cu:89 code=77(cudaErrorIllegalAddress) "cudaMemcpy(h_c,d_plan->c,M*sizeof(CUCPX), cudaMemcpyDeviceToHost)" 
```

`bin/interp2d 1 0 1e3 1e3 1e7 1e-3` works ok
`bin/spread2d` and the 3d ones all work.

ahbarnett:
fixed back in sept 2020.

PR #66, opened 2020-09-24
Update spread/interp only tests

MelodyShih:
This pull request is related to #65 .

Here is a list of changes I make:

- Previously, the spread,interp only functions assume x,y,z,c,fw arrays are on cpu. I changed to use gpu array d_x,d_y,d_z,d_c,d_fw as inputs.
- Add an option "gpu_spreadinterponly" so that we don't allocate/free memory for fw and fwkerhalf arrays when calling ALLOCGPUMEMROY{2,3}D_PLAN and FREEGPUMEMORY{2,3}D.
- Make the spread,interp only test codes be compiled for both precision. 
- Checks for valid values of binsizex, binsizey, binsizez.
- Set binsizex, binsizey, binsizez in the spread,interp{2,3}_test.cu. This fixes the error of running `interp_2d`. binsize{x,y} are set to -1 when calling default options and in the spread,interp only function, SET_BINSIZE is not called.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,8 +108,10 @@ CUFINUFFTOBJS_32=$(CUFINUFFTOBJS_64:%.o=%_32.o)
 %.o: %.cu $(HEADERS)
 	$(NVCC) --device-c -c $(NVCCFLAGS) $(INC) $< -o $@
 
-all: $(BINDIR)/spread2d \
-	$(BINDIR)/interp2d \
+all: $(BINDIR)/spread2d_test \
+    $(BINDIR)/spread2d_test_32 \
+	$(BINDIR)/interp2d_test \
+	$(BINDIR)/interp2d_test_32 \
 	$(BINDIR)/cufinufft2d1_test \
 	$(BINDIR)/cufinufft2d2_test \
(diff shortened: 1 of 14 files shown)


ahbarnett (approved):
Lots of work here - I hope it's because the spread/interp testers will be useful for your paper too?

I tried a few things out - works :+1: 

Certinaly useful to users, since we can now compare single/double speeds:
```
bin/spread2d_test_32 2 0 1e3 1e3 10000 1e7 1e-3
[Method 2] 10000000 NU pts to #1000000 U pts in 0.0269 s (3.72e+08 NU pts/s)
bin/spread2d_test 2 0 1e3 1e3 10000 1e7 1e-3
[Method 2] 10000000 NU pts to #1000000 U pts in 0.0627 s (1.59e+08 NU pts/s)
```
That's on a P4000.  The single is still 6x a modern CPU, not too shabby.
Asking for 1e-9 in double drops it to 4e7 NUpts/s which is CPU speed (and only twice the speed of my i7).

Best, Alex

ahbarnett:
Melody fixed the spread/interp tests back in Sep 2020. not part of make tests, but still useful.

PR #69, opened 2020-10-21
Updates related to spread, interp tests

MelodyShih:
Here is a summarization of what I did in this pull request:

- Add spread/interp to `make check` task
- Modify the timing description of spread, interp only codes (used for benchmarking this part of the codes)
- There are some error I made in the test codes, e.g. not using nf1, nf2, nf3 correctely, memory copy in interp3d_test and also a timing error in spread2d code. I fixed them and also modify some parameter setting so that they are using the default values in cufinufft2,3d1,2 codes.


ahbarnett (approved):
looks good - I trust your improvements to the various binsize and problemsize params.

test/interp3d_test.cu
```diff
@@ -93,8 +93,8 @@ int main(int argc, char* argv[])
 	if(dplan->opts.gpu_method == 1)
 	{
 		dplan->opts.gpu_binsizex=16;
-		dplan->opts.gpu_binsizey=8;
-		dplan->opts.gpu_binsizez=4;
+		dplan->opts.gpu_binsizey=16;
```
janden, line 96: Shouldn't we just be using the default bin sizes?

janden (approved):

test/interp3d_test.cu
MelodyShih: I am using the default bin size is 16x16x2. We have to explicitly specify in spread, interp tests because SETUP_BINSIZE is called inside cufinufft_makeplan and is not defined in the header file.

janden: Ah I see. Maybe we should add a comment so that people aren't confused.

MelodyShih: Yes, I agree. I added the comments in 19bdd30.

PR #72, opened 2020-10-26
fix calculation of kernel width

MelodyShih:
I found that when dividing epsilon by 10, if we don't cast 10.0 to FLT first, kernel width `ns` may not be calculated as expected e.g. for eps=1e-6, ns will be 8 instead of 7.


ahbarnett (approved):
Looks good. This is how I've had it in the CPU FINUFFT for a while now. eps=1e-6 is a borderline case, meaning 1.0001e-6 should give ns=7 but 0.9999e-6 should give ns=8. So it must be that 10.0 was a double const and the conversion to double of spingle-prec eps messed up rounding.

MelodyShih:
I see. I was confused how that happened. Thank you for the explanation.

PR #74, opened 2020-11-08
Fix memory leak

JBlaschke:
The main leak was due to `d_plan->c` being `cudaMalloc`'ed but never `cudaFree`'d if using the python interface. This PR addresses this issue, and adds a few guards against potential memory leaks when the plan gets executed several times without destroying it first.


MelodyShih (approved):
Hi Blaschke, sorry for the late reply. I looked into the changes and they look good. Thanks for finding the memory bugs I made.

JBlaschke:
No worries. Happy to help 😄

Issue #75, opened 2020-12-09
GPU memory leak issue?


penfe:
Dear Yu-hsuan,

I am trying to invoke the cunufft in C++ time and time again, but the monitor for NVIDIA tool [NVTOP](https://github.com/Syllo/nvtop) shows the gpu memory is keep growing.

Would you please kindly review my code below and show me if I have some mistakes in the cunufft API usage? 

~~~c++
#include <cufinufft.h>
#include <complex>
#include <iostream>
#include "unistd.h" // sleep

#define FI_OK 0

void call_finufft(int N1, int N2, int M, int ntransf, float* x, float* y, cuFloatComplex* c, cuFloatComplex* fk)
{
	int maxbatchsize = 0;
	int dim = 2;
	int nmodes[3] = { N1, N2, 1};
	float tol = 1e-6;

	int type = 1;
	int iflag = 1;

	cufinufft_opts opts;
	if (FI_OK != cufinufftf_default_opts(type, dim, &opts))
	{
		std::cout << "err: cufinufft_default_opts" << std::endl;
		return;
	}		
	cufinufftf_plan dplan;
	if (FI_OK != cufinufftf_makeplan(type, dim, nmodes, iflag, ntransf, tol, maxbatchsize, &dplan, NULL))
	{
		std::cout << "err: cufinufftf_makeplan" << std::endl;
		return;
	}
	if (FI_OK != cufinufftf_setpts(M, x, y, NULL, 0, NULL, NULL, NULL, dplan))
	{
		std::cout << "err: cufinufftf_setpts" << std::endl;
		return;
	}
	if (FI_OK != cufinufftf_execute(c, fk, dplan))
	{
		std::cout << "err: cufinufftf_execute" << std::endl;
		return;
	}
	if (FI_OK != cufinufftf_destroy(dplan))
	{
		std::cout << "err: cufinufftf_destroy" << std::endl;
		return;
	}

	std::cout << "ok" << std::endl;
}

void test()
{
	int N1 = 512;
	int N2 = 512;
	int M = 632 * 149;
	int ntransf = 2;

	float *d_x, *d_y;
	CUCPX *d_c, *d_fk;
	cudaMalloc(&d_x,M*sizeof(float));
	cudaMalloc(&d_y,M*sizeof(float));
	cudaMalloc(&d_c,M*ntransf*sizeof(cuFloatComplex));
[...]

PR #76, opened 2020-12-17
Fix memleak

MelodyShih:
This commit fixes memory leak mentioned in #75 . The main errors are missing "`break`" in `ALLOCGPUMEM2D_PLAN` for `gpu_method==2` and the allocation of array `numsubprob` for method 1 in both 2,3D.

### DIFF
--- a/src/memtransfer_wrapper.cu
+++ b/src/memtransfer_wrapper.cu
@@ -28,8 +28,6 @@ int ALLOCGPUMEM2D_PLAN(CUFINUFFT_PLAN d_plan)
 					int numbins[2];
 					numbins[0] = ceil((FLT) nf1/d_plan->opts.gpu_binsizex);
 					numbins[1] = ceil((FLT) nf2/d_plan->opts.gpu_binsizey);
-					checkCudaErrors(cudaMalloc(&d_plan->numsubprob,numbins[0]*
-						numbins[1]*sizeof(int)));
 					checkCudaErrors(cudaMalloc(&d_plan->binsize,numbins[0]*
 						numbins[1]*sizeof(int)));
 					checkCudaErrors(cudaMalloc(&d_plan->binstartpts,numbins[0]*
@@ -51,6 +49,7 @@ int ALLOCGPUMEM2D_PLAN(CUFINUFFT_PLAN d_plan)
 				checkCudaErrors(cudaMalloc(&d_plan->subprobstartpts,
 						(numbins[0]*numbins[1]+1)*sizeof(int)));
 			}
+			break;
 		case 3:
 			{
 				int numbins[2];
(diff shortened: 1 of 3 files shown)


test/cufinufft3d1_test.cu
```diff
@@ -185,7 +185,14 @@ int main(int argc, char* argv[])
 
 	cudaFreeHost(x);
 	cudaFreeHost(y);
+	cudaFreeHost(z);
```
ahbarnett, line 188: ah, yes, the joys of C-style mem mgmt!  But this is only a test code, not the main lib.

src/memtransfer_wrapper.cu
```diff
@@ -51,6 +49,7 @@ int ALLOCGPUMEM2D_PLAN(CUFINUFFT_PLAN d_plan)
 				checkCudaErrors(cudaMalloc(&d_plan->subprobstartpts,
 						(numbins[0]*numbins[1]+1)*sizeof(int)));
 			}
+			break;
```
ahbarnett, line 52: C-style switch is really dumb, having to have break here. It's like the old days of "goto line ...".
Why not replace switch w/ if then else? No breaks required!

ahbarnett:
looks good - I haven't run it - you might want to request Garrett rather than me.

I see numsubprob defined in
``include/cufinufft_eitherprec.h``
Is it the array of how many NU pts go into each subprob? Something like ``counts`` or ``offsets`` in FINUFFT binsort?
Would be good to define it in a comment.


Note Jenkins fails at the py test stage with:
```
   import six
E   ModuleNotFoundError: No module named 'six'
```
Is this unrelated to the memleak fix?

ahbarnett (approved):

MelodyShih:
numsubprob is an array that has number of output bins entries and each entry stores information of how many subproblems to launch for spreading nupts in the output bin. I added some comments in commit c5921c93 .

The Jenkins fail doesn't seem to relate to the change of the codes -- I however don't really know how to fix it:(

garrettwrong:
Hi Melody, I would be happy to look at this tomorrow morning. We did have some trouble with one of our 3D applications in ASPIRE (one in a cg loop), but I never had time to go back to it.  There is a very good chance your patches would help with that case.

Let me look at the missing `six` dependency. Give me a minute on that...

garrettwrong:
I put together the small patch for the missing `six` library #77 .  That CI failure has nothing to do with changes in this PR.  I think some software versions moved since the last time this system was exercised.

garrettwrong (approved):
I checked the code out. Built, ran check, and installed python package on my machine.  Ran some of the tests with Aspire. I didn't intentionally construct an abusive loop, but one of the ones we had in our code seems better now 😅 . (It used to occasionally seg fault, but has not with this code (yet)).

MelodyShih:
@garrettwrong Thank you for taking time reviewing it. Hope that this really fixes the seg fault problem you met. I will merge this pull request.

PR #77, opened 2020-12-18
try adding six here

garrettwrong:
The pycuda package depends should probably include it (since it's trying to use it...), but we can also try to ensure it is installed in this case.

This is read by as part of `setup.py` and should install the dependency when cufinufft is installed, just before the pytest lines..  I think :)


MelodyShih (approved):
Thanks for looking into the issue!

PR #78, opened 2021-01-22
Make Python interface closer to FINUFFT

janden:
Summary:
– `set_pts` infers number of points from input
– `tol` and `ntransforms` are now `eps` and `n_trans`
– order of arguments to `cufinufft` and default values are now the same as Finufft
– `opts` is no longer given as an object but is parsed from the kwargs to `cufinufft`

### DIFF
--- a/python/cufinufft/cufinufft.py
+++ b/python/cufinufft/cufinufft.py
@@ -143,14 +143,13 @@ class cufinufft:
         if ier != 0:
             raise RuntimeError('Error creating plan.')
 
-    def set_pts(self, M, kx, ky=None, kz=None):
+    def set_pts(self, kx, ky=None, kz=None):
         """
         Sets non uniform points of the correct dtype.
 
         Note kx, ky, kz are required for 1, 2, and 3
         dimensional cases respectively.
 
-        :param M: Number of points
         :param kx: Array of x points.
         :param ky: Array of y points.
         :param kz: Array of z points.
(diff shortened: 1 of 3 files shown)


python/cufinufft/cufinufft.py
```diff
@@ -60,12 +60,12 @@ def __init__(self, nufft_type, modes, isign, tol,
...
-        self.opts = opts
+        if isign is None:
+            if nufft_type == 2:
+                isign = -1
+            else:
+                isign = +1
+
 
         # Setup type bound methods
         self.dtype = np.dtype(dtype)
```
ahbarnett, line 71: in FINUFFT
https://github.com/flatironinstitute/finufft/blob/b1a1f860fa974ae3295cd59c8cfe5bd2286e5f73/python/finufft/_interfaces.py#L97

you use more complicated way to check dtype in kwargs... but I guess it's the same thing.

ahbarnett (approved):
new interface looks good to me. have not seen the device_id aspect yet.

MelodyShih:
I tested the new interface on Flatiron's machine -- works good.

python/cufinufft/cufinufft.py
janden: I don't think what we're doing in FINUFFT to get the dtype is necessary if we provide it as a regular argument. It will be removed from the `kwargs` dict and processed just like the other arguments. I'm guessing that code is a holdover from some earlier attempt at parsing the opts.

Issue #79, opened 2021-01-26
set_pts memory leak


janden:
I want to point out though that using `cufinufft` that way presupposes the user is carefully managing memory anyway, as  `CUFINUFFT_SETPTS`  always mallocs (which is assumed to be freed with the plan in `CUFINUFFT_DESTROY`) => so by setting new points without first destroying+creating a plan will cause a memory leak (this might be a good discussion in a different issue @MelodyShih @ahbarnett). 

_Originally posted by @JBlaschke in https://github.com/flatironinstitute/cufinufft/issues/71#issuecomment-767194388_

MelodyShih:
Hi Joakim, I agree with the issue. I think adding a check to see if the arrays have already be allocated or not in function `ALLOCGPUMEM2,3D_NUPTS` before malloc them will fix it. I will create a pull request to fix it.

janden:
Fixed by #80.

PR #80, opened 2021-01-27
Fix memory leak

MelodyShih:
These commits fix the mem leak issue #79 . Here is the output of `cuda-memcheck`:

cuda-memcheck --leak-check full --tool memcheck --show-backtrace host bin/cufinufft2d1nupts_test_32 1 128 128
========= CUDA-MEMCHECK
[time  ] dummy warmup call to CUFFT	 587 s
[time  ] cufinufft plan:		 0.0011 s
[time  ] cufinufft setNUpts (set 1):	 0.0482 s
[time  ] cufinufft exec (set 1):	 0.0269 s
[time  ] cufinufft setNUpts (set 2):	 0.052 s
[time  ] cufinufft exec (set 2):	 0.0256 s
[time  ] cufinufft destroy:		 0.00535 s
[Method 1] (16384+32768) NU pts to 16384 U pts in 0.159 s:      3.09e+05 NU pts/s
					(exec-only thoughput: 9.36e+05 NU pts/s)
[gpu   ] one mode: rel err in F[47,33] is 9.16e-07 (set 1)
[gpu   ] one mode: rel err in F[47,33] is 7.82e-07 (set 2)
========= LEAK SUMMARY: 0 bytes leaked in 0 allocations
========= ERROR SUMMARY: 0 errors

### DIFF
--- a/src/2d/spread2d_wrapper.cu
+++ b/src/2d/spread2d_wrapper.cu
@@ -586,6 +586,7 @@ int CUSPREAD2D_SUBPROB_PROP(int nf1, int nf2, int M, CUFINUFFT_PLAN d_plan)
 	MapBintoSubProb_2d<<<(numbins[0]*numbins[1]+1024-1)/1024, 1024>>>(
 			d_subprob_to_bin,d_subprobstartpts,d_numsubprob,numbins[0]*numbins[1]);
 	assert(d_subprob_to_bin != NULL);
+        if (d_plan->subprob_to_bin != NULL) cudaFree(d_plan->subprob_to_bin);
 	d_plan->subprob_to_bin = d_subprob_to_bin;
 	assert(d_plan->subprob_to_bin != NULL);
 	d_plan->totalnumsubprob = totalnumsubprob;
--- a/test/cufinufft2d1nupts_test.cu
+++ b/test/cufinufft2d1nupts_test.cu
@@ -1,0 +1,246 @@
+#include <iostream>
+#include <iomanip>
+#include <math.h>
+#include <helper_cuda.h>
+#include <complex>
+
+#include <cufinufft_eitherprec.h>
+
+#include "../contrib/utils.h"
+
+using namespace std;
+
+int main(int argc, char* argv[])
+{
+	int N1, N2, M1, M2, N;
+	if (argc<4) {
+		fprintf(stderr,
+			"Usage: cufinufft2d1nupts_test method N1 N2\n"
+			"Arguments:\n"
+			"  method: One of\n"
+			"    1: nupts driven,\n"
+			"    2: sub-problem, or\n"
+			"    3: sub-problem with Paul's idea.\n");
+		return 1;
+	}
+	int method;
+	sscanf(argv[1],"%d",&method);
+
+	N1 = 100;
+	N2 = 100;
+	N = N1*N2;
+	M1 = N1*N2;
+	M2 = 2*N1*N2;
+
+	FLT tol=1e-5;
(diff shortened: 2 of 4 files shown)


src/2d/spread2d_wrapper.cu
```diff
@@ -586,6 +586,7 @@ int CUSPREAD2D_SUBPROB_PROP(int nf1, int nf2, int M, CUFINUFFT_PLAN d_plan)
 	MapBintoSubProb_2d<<<(numbins[0]*numbins[1]+1024-1)/1024, 1024>>>(
 			d_subprob_to_bin,d_subprobstartpts,d_numsubprob,numbins[0]*numbins[1]);
 	assert(d_subprob_to_bin != NULL);
+        if (d_plan->subprob_to_bin != NULL) cudaFree(d_plan->subprob_to_bin);
```
janden, line 589: Shouldn't the assert above catch the NULL? I don't see why the if is needed.

test/cufinufft2d1nupts_test.cu
```diff
@@ -252,5 +252,9 @@ int main(int argc, char* argv[])
 	cudaFree(d_y2);
 	cudaFree(d_c2);
 	cudaFree(d_fk2);
+	
+	// for cuda-memcheck to work
+	cudaDeviceReset();
```
janden, line 257: Is this something we should be doing in the other tests as well?

janden (changes requested):
This looks good to me. The only thing I'm not sure about is whether the new unit tests needs all those parameters. It's just there to check whether we can call `set_pts` twice without incurring a memory leak. We're not testing a lot of different parameters here. On a related note, would it make sense to run a memcheck on all of the tests as a standard part of the tests? Is there some dependency that's needed for that?

Seems to be some internal Jenkins issue that's holding this up. I've tried restarting it a few times but it keeps failing. If it doesn't resolve itself I'll get in touch with the admins.

src/2d/spread2d_wrapper.cu
MelodyShih: The if statement is checking the `d_plan->subprob_to_bin` pointer. The next line (not showing here), we assign `d_subprob_to_bin` to  `d_plan->subprob_to_bin`.

test/cufinufft2d1nupts_test.cu
MelodyShih: I don't think it is necessary. According to this stack overflow post: https://stackoverflow.com/questions/36012289/what-is-the-role-of-cudadevicereset-in-cuda, the call of the function is not considered mandatory. But as the discussions in the answer suggest, it is required for using tools like `cuda-memcheck` or using the profilers.

MelodyShih:
I removed the unnecessary arguments in the latest commit. 
There is no other dependency of using `cuda-memcheck` except the call of `cudaDeviceReset()`. But since it takes time to check the memory leak (10 minutes for a run), I don't think we want to add it as the standard part of the tests (`make check`? ).

test/cufinufft2d1nupts_test.cu
janden: Ok sounds good.

janden:
Ok that makes sense. It would be nice if we could do something like `cuda-memcheck make check` or `make memcheck` though. Not now, but something to think about later.

janden (approved):

MelodyShih:
I like the `make memcheck` idea! We can then put the essential test codes there instead of memchecking all of the test codes we have now.

PR #82, opened 2021-01-28
Interface cleanup

janden:
Some things I forgot to fix last week.

### DIFF
--- a/python/cufinufft/cufinufft.py
+++ b/python/cufinufft/cufinufft.py
@@ -104,7 +104,7 @@ class cufinufft:
         self.modes = (c_int * 3)(*modes)
 
         # Get the default option values.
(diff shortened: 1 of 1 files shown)


garrettwrong (approved):
I think this is fine, but we need to remember to remind users (you and me :) ) to update any use of default_opts. I think there is the 3d case in ASPIRE where we get the default opts then override the gpu method (to get around the shmem limit of default algo in doubles).

janden:
Right. I can take care of that once 1.2 is up on PyPI. In the meantime, I'll put in a PR to pin the version to 1.1.

garrettwrong:
Sure if you want. Otherwise I'll do it tomorrow morning. No trouble... was busy with the openacc workshop today. We'll be okay since this isn't breaking until 1.2 is pushed up.

PR #83, opened 2021-01-28
Fix C++ examples and add Python examples

janden:
Couldn't get the C++ examples to run at first since they seem to be using the old (struct pointer) interface. It would be good to somehow include the examples in the CI so that we know they're correct at all times.

The Python examples follow the C++ examples more or less, I just changed the parameters slightly.

### DIFF
--- a/examples/example2d2many.py
+++ b/examples/example2d2many.py
@@ -1,0 +1,60 @@
+"""
+Demonstrate the type 2 NUFFT using cuFINUFFT
+"""
+
+import numpy as np
+
+import pycuda.autoinit
+from pycuda.gpuarray import GPUArray, to_gpu
+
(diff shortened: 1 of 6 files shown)


examples/example2d2many.py
```diff
@@ -0,0 +1,62 @@
+"""
+Demonstrate the type 22NUFFT using cuFINUFFT
```
garrettwrong, line 2: 22 typo

garrettwrong (approved):
I must have missed changing the plan pointer dereference level change in the example codes, thanks for catching that.

examples/example2d2many.py
janden: Huh that's a weird one.

PR #86, opened 2021-01-29
remove -g flag

MelodyShih:
Here is one minor change we discussed to do -- remove the host compiler option `-g`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ NVARCH ?= -arch=sm_70 \
 	  -gencode=arch=compute_75,code=sm_75 \
 	  -gencode=arch=compute_75,code=compute_75
 
-CFLAGS    ?= -fPIC -O3 -funroll-loops -march=native -g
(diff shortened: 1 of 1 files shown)


garrettwrong:
i have to ask why? In the old days, dark ages, before I was born, this was really a thing, but I've never personally seen it be a performance issue since I've been on the scene. I am open to being wrong on this point, but I need proof :) It does in fact make the binaries larger, but time to solution isn't usually a measurable issue. Personally I would always tell students to use debugging symbols, so I'm sort of antithetical on this point... debugging under optimization  is one of the things the gnu folks gifted us.... This almost breaks my heart. Convert me <3

janden:
I see. I'd argue it's a matter of simplicity. We don't need it, so why add the noise? (And it is noise for those of us who don't necessarily know that adding debug flags *doesn't* slow you down.)

ahbarnett:
Agreed; any competent user can add -g.

ahbarnett (approved):
Yes.

PR #94, opened 2021-02-06
Updates for CentOS 6 deprecation

### LINKED ISSUE #92: Fixup Centos6 repo references for manylinux2010 images
Centos6 was end of lifed, and so some references to repos no longer resolve.  This has the effect of breaking my local attempts at building the docker images (that are used to make the 2010 wheels, CUDA 10).  It will probably effect other developers if not already..

I think to fix it just requires pointing to the vault copy of centos6. I'll look monday.

Moving forward the project can use manylinux2014, which is based on centos 7 and cuda 11. Making a separate issue about that.

```
...
Step 7/21 : RUN yum install -y cuda-cudart-$CUDA_PKG_VERSION cuda-compat-10-1 &&     ln -s cuda-10.1 /usr/local/cuda &&     rm -rf /var/cache/yum/*
 ---> Running in 44a0be3e52b1
Loaded plugins: fastestmirror, ovl
Setting up Install Process
Determining fastest mirrors
Error: Cannot find a valid baseurl for repo: base
YumRepo Error: All mirror URLs are not using ftp, http[s] or file.
...
```

garrettwrong:
Redirects yum links to CentOS 6 vault.

This fixed up `ci/docker/cuda10.1/Dockerfile-x86_64` for me.

Closes #92

### DIFF
--- a/ci/docker/cuda10.1/CentOS-SCLo-scl-rh.repo
+++ b/ci/docker/cuda10.1/CentOS-SCLo-scl-rh.repo
@@ -1,0 +1,33 @@
+# CentOS-SCLo-rh.repo
+#
+# Please see http://wiki.centos.org/SpecialInterestGroup/SCLo for more
+# information
+
+[centos-sclo-rh]
+name=CentOS-6 - SCLo rh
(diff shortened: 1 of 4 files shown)


janden (approved):
LGTM

PR #96, opened 2021-02-07
Store GPUArray as variables

### LINKED ISSUE #85: Inaccurate type 1 transform in Python
When running the type 1 transform in Python with a small number of nonuniform points, the output seems to be garbage (i.e., order one error). Sometimes it's zero, sometimes not. Initializing the output array with a fixed value shows that something the array is being written to. I've not been able to reproduce this in the C interface, so it's likely this has something to do with the Python binding.

```
import numpy as np
import pycuda.autoinit
from pycuda.gpuarray import GPUArray, to_gpu

N1, N2 = 256, 256
M = 10
dtype = np.float32
complex_dtype = np.complex64

kx = np.random.uniform(-np.pi, np.pi, size=M)
ky = np.random.uniform(-np.pi, np.pi, size=M)

c = (np.random.standard_normal(M)
     + 1j * np.random.standard_normal(M))

kx = kx.astype(dtype)
ky = ky.astype(dtype)
c = c.astype(complex_dtype)

fk_gpu = to_gpu(np.full(fill_value=-123, shape=(N1, N2),
    dtype=complex_dtype))

plan = cufinufft(1, (N1, N2), dtype=dtype)
plan.set_pts(to_gpu(kx), to_gpu(ky))

plan.execute(to_gpu(c), fk_gpu)

fk = fk_gpu.get()

nt1 = int(0.37 * N1)
nt2 = int(0.26 * N2)

x, y = nt1 - N1 // 2, nt2 - N2 // 2
fk_true = np.sum(c * np.exp(1j * (x * kx + y * ky)))

print(f"fk={fk[nt1, nt2]}, fk_true={fk_true}")

err = np.abs(fk[nt1, nt2] - fk_true)
rel_err = err / np.max(np.abs(fk))

print(f"err={err}, rel_err={rel_err}")
```

MelodyShih:
To fix #85.

### DIFF
--- a/examples/example2d1many.py
+++ b/examples/example2d1many.py
@@ -30,16 +30,21 @@ kx = kx.astype(dtype)
 ky = ky.astype(dtype)
 c = c.astype(complex_dtype)
 
+# Transfer memory from host to device
+kx_gpu = to_gpu(kx)
+ky_gpu = to_gpu(ky)
+c_gpu  = to_gpu(c)
+
 # Allocate memory for the uniform grid on the GPU.
 fk_gpu = GPUArray((n_transf, N1, N2), dtype=complex_dtype)
 
 # Initialize the plan and set the points.
 plan = cufinufft(1, (N1, N2), n_transf, eps=eps, dtype=dtype)
-plan.set_pts(to_gpu(kx), to_gpu(ky))
+plan.set_pts(kx_gpu, ky_gpu)
 
 # Execute the plan, reading from the strengths array c and storing the
 # result in fk_gpu.
-plan.execute(to_gpu(c), fk_gpu)
+plan.execute(c_gpu, fk_gpu)
 
 # Retreive the result from the GPU.
 fk = fk_gpu.get()
--- a/examples/example2d2many.py
+++ b/examples/example2d2many.py
@@ -30,16 +30,21 @@ kx = kx.astype(dtype)
 ky = ky.astype(dtype)
 fk = fk.astype(complex_dtype)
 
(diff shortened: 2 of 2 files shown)


ahbarnett:
Looks sensible (if a bit annoying, because our code example in the paper uses to_gpu() inside the setpts func call... but the reader will just have to read the actual example in the repo).

Andreas' [documentation for pycuda](https://documen.tician.de/pycuda/) doesn't make this clear, just saying things like:
PyCUDA will automatically infer what cleanup is necessary and do it for you.

I can't get py+cuda running, so I'll leave someone else to test this. Thanks, Alex

garrettwrong:
I think the reference count to the object isn't incrementing and decrementing as desired when using the raw pointer exposed from `.ptr` (It couldn't know what to do anyway).

So in the first implementation that is probably why the array objects are getting cleaned up too early... they are not used elsewhere.  The second example keep them in scope longer

One way direct way might be to use `Py_INCREF` and `Py_DECREF`.  A higher level way would be to save the array objects as an instance attribute, or in a list that belongs to the instance,  for as long as we care about the instance. I did that for aspire, which is probably why I haven't run into this...

https://github.com/ComputationalCryoEM/ASPIRE-Python/blob/master/src/aspire/nufft/cufinufft.py#L81

If you want I can put that together tomorrow. Should be very minor patch and you may keep your examples as they were... I can probably also take that line out of aspire then :)

MelodyShih:
Use the solution proposed in #97 instead.

PR #97, opened 2021-02-08
Hold a reference to set_pts gpuarrays in the class

### LINKED ISSUE #85: Inaccurate type 1 transform in Python

garrettwrong:
@MelodyShih , try this for #85 . I think this will let you keep your examples as they were. This should keep python from cleaning up the set_pts arrays before the cufinufft class instance is cleaned up. 

Closes #85

### DIFF
--- a/python/cufinufft/cufinufft.py
+++ b/python/cufinufft/cufinufft.py
@@ -119,6 +119,10 @@ class cufinufft:
         # Initialize the plan.
         self._plan()
 
+        # Initialize a list for references to objects
+        #   we want to keep around for life of instance.
+        self.references = []
+
     @staticmethod
     def _default_opts(nufft_type, dim):
         """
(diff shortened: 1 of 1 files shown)


MelodyShih (approved):
@garrettwrong, thanks for the new solution! I tested and confirmed that it fixes #85. 
I will close #96.

janden (approved):
Works great for me. Thanks!

Issue #93, opened 2021-02-05
Move towards manylinux2014 wheels


garrettwrong:
Following #92 , it is time to move to newer OS and CUDA basis for the python wheels.

- [x] Confirm manylinux2014 ci image builds (`make` inside the container)
- [x] Confirm `make check` inside the container succeeds on my machine
- [ ] Confirm container build and check on a second machine
- [x] Confirm build wheel script runs
- [x] Check each wheel output 
- [x] Check if python 3.5 is deprecated; if so purge it
- [x] Update the ci/readme to be a little more scriptable (vars for version etc). Perhaps make it a real script and just have CI call the script?.. (probably)
- [x] Consider what the `next thing` after 2014 will be (manylinux, OS, CUDA version) so project is prepared again.

garrettwrong:
Confirmed Python 3.5 is end of life (https://www.python.org/dev/peps/pep-0478/).  It appears `pip` itself will be dropping support for it based on this note I found testing the 2014 3.5 wheel:

```
DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.
```

I'll put together a small change set removing 3.5 from the build script along with any changes needed for 2014 wheel support (had some trouble with 3.7, which I think is why I didn't use these wheels in the summer; memories starting to come back to me).

As far as the `next thing`, it doesn't appear to have been decided in a PEP yet.  2014 is claimed to go through 2024. I can ask around, but I don't know there is much else we can prepare for yet.

ahbarnett:
don't forget to update README.md to state what the new manylinux/CUDA versions allowed are. Thx -A

garrettwrong:
Janden has independently constructed 2014 wheels for testing, and I have confirmed those install and pass unit tests on TigerGPU for python 3.6, 3.7, 3.8, 3.9.

Other items have already been integrated into version 1.2.

Closing.

PR #98, opened 2021-02-08
Some updates for 2014 wheels, and general distribution.

garrettwrong:
Towards #93 

I made a small change to successfully build manylinux2014 wheels for python 3.6 3.7 3.8 and 3.9.

Tested each of those wheels locally. It would be good to validate on a totally unrelated second machine.

Updated the ci readme into a script so in the future its just changing a couple variables.

Hope that makes things easier.


janden:
So if I understand correctly, we'd only be producing manylinux2014 wheels when this is merged? Even though manylinux2010 is getting a bit old, it might be a good idea to ship both for now. From what I understand, pip will install whichever version is compatible with the system if more than one binary wheels are available (see https://www.python.org/dev/peps/pep-0425/#id1).

garrettwrong:
> So if I understand correctly, we'd only be producing manylinux2014 wheels when this is merged? 

The purpose of the PR was more that I needed to make a small patch to the 2014 dockerfile so 2014 wheels would build.  The C library was fine, but python3.7 had trouble finding a library (fixed).

Not necessarily "only". This adds facility to make the 2014 wheels.  Nothing about this PR limits 2010, though I did update the helper default to the newer target.  If you want to build 2010 wheels, change the variables in the helper script (cuda_version=10.1, manylinux_version=manylinux2010), or run the commands manually. (You'll probably need the 2010 changes merged upstream for 2010).

> Even though manylinux2010 is getting a bit old, it might be a good idea to ship both for now. From what I understand, pip will install whichever version is compatible with the system if more than one binary wheels are available (see https://www.python.org/dev/peps/pep-0425/#id1).

Yes, I suggested in the emails with Alex to ship both 2010 and 2014 this release if we can. Both targets needed a PR to work. 2010 because centos6 EOL, and 2014 because of a library path issue.

I don't think that doc link is actually related to what happens for manylinux stuff... They have their own peps which came later (they weren't even introduced at the time that doc was accepted), but the general sentiment is the same, sure.

Also this doesn't update the Jenkins CI stuff.

janden:
> The purpose of the PR was more that I needed to make a small patch to the 2014 dockerfile so 2014 wheels would build. The C library was fine, but python3.7 had trouble finding a library (fixed).

Ok got it.

> Not necessarily "only". This adds facility to make the 2014 wheels. Nothing about this PR limits 2010, though I did update the helper default to the newer target. If you want to build 2010 wheels, change the variables in the helper script (cuda_version=10.1, manylinux_version=manylinux2010), or run the commands manually. (You'll probably need the 2010 changes merged upstream for 2010).

Ok.

> Yes, I suggested in the emails with Alex to ship both 2010 and 2014 this release if we can. Both targets needed a PR to work. 2010 because centos6 EOL, and 2014 because of a library path issue.

Good.

> I don't think that doc link is actually related to what happens for manylinux stuff... They have their own peps which came later (they weren't even introduced at the time that doc was accepted), but the general sentiment is the same, sure.

Right I was trying to dig up a more relevant link but wasn't able to find it.

> Also this doesn't update the Jenkins CI stuff.

Ok.

janden (approved):

PR #101, opened 2021-02-17
Bump version to 1.2

janden:


janden:
@garrettwrong I assumed that `distribution_helper` should be included in the bumpversion config. Please correct me if this omission was intentional.

garrettwrong (approved):
re `distribution_helper` I forgot about it, good catch.  I see you got Alex's readme change too.  Otherwise seems like it worked.  ~I think the only thing left is to git push the tag...~ (nm see it)

PR #103, opened 2021-02-18
Fix build wheels

janden:
Add a `manylinux` target that downgrades the instruction set to SSE4 to increase compatibility. (We're (hopefully) not reliant on AVX(2) instructions for our performance, so this is not a big restriction.) We then use this target in the wheel-building scripts to make sure our binary wheels will run on older CPUs.

Also upgrades pip to remove those warnings.

### DIFF
--- a/ci/docker/cuda11.0/Dockerfile-x86_64
+++ b/ci/docker/cuda11.0/Dockerfile-x86_64
@@ -75,7 +75,7 @@ ENV LIBRARY_PATH /usr/local/cuda/lib64/stubs
 # assume we are building container in the root of the git repo...
 COPY . /io
 WORKDIR /io
-RUN make
+RUN make target=manylinux2010
 # And we need to pack it in our LD path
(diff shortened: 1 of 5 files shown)


ci/docker/cuda11.0/Dockerfile-x86_64
```diff
@@ -75,7 +75,7 @@ ENV LIBRARY_PATH /usr/local/cuda/lib64/stubs
 # assume we are building container in the root of the git repo...
 COPY . /io
 WORKDIR /io
-RUN make
+RUN make target=manylinux2010
```
garrettwrong, line 78: Did you just mean `manylinux` here?

garrettwrong (changes requested):

garrettwrong (approved):

ci/docker/cuda11.0/Dockerfile-x86_64
janden: Right. This is why the 2014 wheels ended up with AVX2…

PR #104, opened 2021-02-19
Update readme to differentiate CUDA 10 release and test release for C…

garrettwrong:
…UDA 11.

If we want to remove reference entirely just let me know and I'll cut it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ General Python users, or Python software packages which would like to automatica
 depend on cufinufft using `setuptools` may use a precompiled binary distribution.
 This totally avoids installing from source and managing libraries for supported systems.
 
-Binary distributions are specific to both hardware and software. We currently provide binary wheels targeting Linux systems covered by `manylinux2010` for CUDA 10 and `manylinux2014` for CUDA 11 systems with compatible GPUS. These are currently the most common arrangements.  If you have such a system, you may run:
+Binary distributions are specific to both hardware and software. We currently provide binary wheels targeting Linux systems covered by `manylinux2010` for CUDA 10 forward with compatible GPUs.  If you have such a system, you may run:
 
 `pip install cufinufft`
 


README.md
```diff
@@ -110,10 +110,16 @@ General Python users, or Python software packages which would like to automatica
 depend on cufinufft using `setuptools` may use a precompiled binary distribution.
 This totally avoids installing from source and managing libraries for supported systems.
 
-Binary distributions are specific to both hardware and software. We currently provide binary wheels targeting Linux systems covered by `manylinux2010` for CUDA 10 and `manylinux2014` for CUDA 11 systems with compatible GPUS. These are currently the most common arrangements.  If you have such a system, you may run:
+Binary distributions are specific to both hardware and software. We currently provide binary wheels targeting Linux systems covered by `manylinux2010` for CUDA 10 forward with compatible GPUs.  If you have such a system, you may run:
 
 `pip install cufinufft`
 
+We have a similar `manylinux2014` wheel based on CUDA 11 for testing with the latest Ampere GPUs.
```
janden, line 117: I'm not sure about this way of distributing the 2014 wheels. From what I understand, Test PyPI is only for testing out distribution processes, not as an alternative distribution channel to end users. For now, I think it's simpler to just have them contact us for the 2014 wheels (alternatively publish them as “Releases” on GH, although I'm less familiar with this).

Another possible approach (if we choose to maintain multiple CUDA versions, which I'm not sure we will) is to use `pip3 install -f` and provide a URL where we publish links to the wheels, like PyTorch: https://download.pytorch.org/whl/torch_stable.html

janden (changes requested):
I'd rather remove the

README.md
garrettwrong: k

janden (approved):

PR #106, opened 2021-02-22
Add manifest

janden:
This makes sure that `requirements.txt` and the Python `README.md` are
included in the source distribution.

Also fixes `os.path.join` in setup.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include python/cufinufft/README.md
+include python/cufinufft/requirements.txt
--- a/setup.py
+++ b/setup.py
@@ -8,11 +8,11 @@ from setuptools import setup, Extension
 # Description
 DESCRIPTION = "Non-uniform fast Fourier transforms on the GPU"
 
-with open('python/cufinufft/README.md', encoding='utf8') as fh:
+with open(os.path.join('python', 'cufinufft', 'README.md'), encoding='utf8') as fh:
     LONG_DESCRIPTION = fh.read()
 
 # Parse the requirements
-with open(os.path.join('python/cufinufft', 'requirements.txt'), 'r') as fh:
+with open(os.path.join('python', 'cufinufft', 'requirements.txt'), 'r') as fh:
     requirements = [item.strip() for item in fh.readlines()]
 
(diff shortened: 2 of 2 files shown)


garrettwrong (approved):
Oh, sorry I missed this.  I've never actually tried to install just the wrapper from pypi (since cufinufft lib doesnt actually have an official system level `make install` this feels strange). Locally I just did and confirmed the manifest issue. 

I was able to use this branch to generate an sdist. Install required installing my OS's python3-devel package among other tasks. Then with the library in LD_LIBRARY_PATH and LIBRARY_PATH, install succeeded. This is potentially different for different systems...

I was also able to install the sdist under anaconda, but I again needed to have the LIBRARY_PATH set (or the library can be copied into anaconda's lib area would do).

Overall, I would recommend users in this situation build from source outside of pip. Installing from the repo is slightly more robust. It isn't necessary to ship the sdist. Might be worth considering if you really want to ship that.

janden:
> Oh, sorry I missed this. I've never actually tried to install just the wrapper from pypi (since cufinufft lib doesnt actually have an official system level make install this feels strange). Locally I just did and confirmed the manifest issue.

Yes this happened when using an old version of pip (I think) which didn't support manylinux2010 so it tried the sdist. Agree that it's a strange use case.

> I was able to use this branch to generate an sdist. Install required installing my OS's python3-devel package among other tasks. Then with the library in LD_LIBRARY_PATH and LIBRARY_PATH, install succeeded. This is potentially different for different systems...

Same here.

> Overall, I would recommend users in this situation build from source outside of pip. Installing from the repo is slightly more robust. It isn't necessary to ship the sdist. Might be worth considering if you really want to ship that.

Ok, I wasn't sure why the sdist was in there, but you're right, it doesn't make much sense since a user would have to download the source and compile themselves and could just take the next step and run `make python` in that case. It's just going to confuse users to have the sdist in there.

I'm going to remove the sdist from the wheel script and will only upload the binary wheels for 1.2. I'm guessing we should remove previous versions of the sdist as well since that's likely to be even more confusing (“Why is there an sdist for 1.1 but not for 1.2?”).

PR #108, opened 2021-03-11
Fix python doc of isign variable

MelodyShih:
As titled.


garrettwrong (approved):
This might have been my typo; I don't recall, but it is something I would do :).

Thank you for fixing.

Issue #112, opened 2021-07-07
[bug report] incorrect 2d type 2 nufft result


tianrluo:
(Description edited to make assessment easier)

It is likely an error of rounding or mod operations related to coordinates.
Play with the `x_ofst`, `y_ofst` below to see the error shows/disappears.

The following lines can reproduce it:

```python
import numpy as np
import finufft, cufinufft, pycuda
import pycuda.autoinit

from pycuda import gpuarray

sx, sy = 4, 4
x_ofst, y_ofst = 0.00, 0.00  # change this to reproduce the error

x, y = np.mgrid[-sx//2:sx//2:1., -sy//2:sy//2:1.]
x, y = x+x_ofst, y+y_ofst
xr, yr = x.reshape(-1)/sx * 2 * np.pi, y.reshape(-1)/sy * 2 * np.pi

p = np.zeros((sx, sy), dtype=np.complex128)
p[0][0] = 1

xr_cu, yr_cu, p_cu = gpuarray.to_gpu(xr), gpuarray.to_gpu(yr), gpuarray.to_gpu(p)

plan = finufft.Plan(2, (sx, sy), 1, eps=1e-12)
plan.setpts(xr.reshape(-1), yr.reshape(-1))

plan_cu = cufinufft.cufinufft(2, (sx, sy), eps=1e-12, dtype=np.float64)
plan_cu.set_pts(xr_cu, yr_cu)

kp = plan.execute(p)

kp_cu = gpuarray.GPUArray((sx*sy,), dtype=np.complex128)
plan_cu.execute(kp_cu, p_cu)
kp_cu_np = kp_cu.get()

kp, kp_cu_np = kp.reshape((sx,sy)), kp_cu_np.reshape((sx,sy))

print(kp.real)
print(kp_cu_np.real)
```

and the outputs are:
```python
# kp.real, correct
array([[ 1. -1.  1. -1.]
 [-1.  1. -1.  1.]
 [ 1. -1.  1. -1.]
 [-1.  1. -1.  1.]])

# kp_cu_np, incorrect
array([[ 1.  0.  1. nan]
 [nan  0.  0.  0.]
 [ 1.  0.  0. nan]
 [ 0. nan nan nan]])
```

MelodyShih:
Hi tianrluo, thank you for reporting this. 
I am able to reproduce the error. I will investigate it and update with you what went wrong.

tianrluo:
Thanks!
Looking forward to a fix!

[...]

PR #113, opened 2021-07-07
Add checks to ensure distance is strictly less than half of kernel width

MelodyShih:
This is related to #112 .

The error happens when nonuniform points lie on the fine grid. 
The distance between the point and the grid point are exactly `kernelwidth/2` and cause 
`exp(es_beta * (sqrt(1.0 - es_c*x*x)))` evaluates to `nan`. 

I added checks to enforce this does not happen.

### DIFF
--- a/src/2d/spreadinterp2d.cu
+++ b/src/2d/spreadinterp2d.cu
@@ -13,7 +13,7 @@ using namespace std;
 #define MAXBINSIZE 1024
 
 static __forceinline__ __device__
-FLT evaluate_kernel(FLT x, FLT es_c, FLT es_beta)
+FLT evaluate_kernel(FLT x, FLT es_c, FLT es_beta, int ns)
 	/* ES ("exp sqrt") kernel evaluation at single real argument:
 	   phi(x) = exp(beta.sqrt(1 - (2x/n_s)^2)),    for |x| < nspread/2
 	   related to an asymptotic approximation to the Kaiser--Bessel, itself an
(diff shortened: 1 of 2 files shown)


ahbarnett:
Yes, this was needed in FINUFFT too, when the plain exp(sqrt(..)) kernel used:
https://github.com/flatironinstitute/finufft/blob/adcb1d25aa8f3b26aed1956a9dccc12bd8a78eea/src/spreadinterp.cpp#L639
With the vectorized Horner eval, there are no NaNs, so no such check was needed.

Note that your PR has an awful lot of repetition of the check. Could it be rationalized so that the kernel eval and associated check is done in one place only?

You might want to check that tweaks to rounding and wrapping in FINUFFT over the last year have been brought into cuFINUFFT. I can list commits if you need, but have a look in spreadinterp.cpp, basically.
Best, Alex

src/2d/spreadinterp2d.cu
```diff
@@ -565,13 +570,13 @@ void Interp_2d_NUptsdriven(FLT *x, FLT *y, CUCPX *c, CUCPX *fw, int M, const int
 		cnow.y = 0.0;
 		for(int yy=ystart; yy<=yend; yy++){
 			FLT disy=abs(y_rescaled-yy);
-			FLT kervalue2 = evaluate_kernel(disy, es_c, es_beta);
+			FLT kervalue2 = disy<ns/2.0 ? evaluate_kernel(disy, es_c, es_beta):0;
```
ahbarnett, line 573: Hi, while this works, I suggest you incorporate this check into evaluate_kernel. This requires adding an extra arg, ns. But then you won't have the same check repeated 10 times as below, which is much harder to maintain...

ahbarnett (changes requested):
Hi Melody - sorry, if you have time, could you tweak this as in my code comment, for cleaner non-repetitive code?  Thanks, Alex

MelodyShih:
Hi Alex, sure, sounds better to do what you suggest. I will update with a new commit. Best regards, Melody

ahbarnett (approved):

PR #116, opened 2021-11-12
Add support for AMD GPUs via HIP

elliottslaughter:
This is work in progress branch to add support for AMD GPUs via [HIP](https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-GUIDE.html), a very nearly CUDA-compatible API by AMD. While this isn't ready to go yet, it's far enough along that I wanted to get feedback on the approach I'm taking.

The basic idea of HIP is to provide CUDA-compatible APIs that are identical except for naming. In most cases you can find-and-replace `cuda` for `hip` and things will work.

Rather than literally doing that transformation to the code (the way AMD seems to want you to do it), I'm using an approach with a header wrapper around `cuda.h` / `hip/hip_runtime.h`. Then I add some translations with `#define cudaX hipX` for each `X` API call used in the code. The advantage of going this route is that the vast majority of the code does not need to change at all, at least for a functionally correct version. I haven't looked at performance at all yet, and there may be more work to do there.

I also made a few minor changes to the build system, mostly to provide enough configuration flexibility so that site files are sufficient to build against HIP. I provide an example site in `sites/make.inc.olcf_spock` for [Spock](https://docs.olcf.ornl.gov/systems/spock_quick_start_guide.html).

Status:

 * With this configuration, I'm able to build and run on [Spock](https://docs.olcf.ornl.gov/systems/spock_quick_start_guide.html). I am testing against ROCm 4.2.0.
 * Running `make site=olcf_spock check` currently gives me failures in the following tests. All other tests are passing.

   ```
   bin/cufinufft2d2api_test
   bin/cufinufft2d2api_test_32
   bin/cufinufft3d1_test 4 15 15 15 2048 1e-3
   bin/cufinufft3d1_test 4 15 15 15
   bin/cufinufft3d1_test_32 4 15 15 15 2048 1e-3
   bin/cufinufft3d1_test_32 4 15 15 15
   $(BINDIR)/example2d1many
   $(BINDIR)/example2d2many
   ```

   The first and last tests seem to be some sort of an issue with the shared library (`"Cannot find Symbol"`). The others I'm not sure about.

I'd appreciate a review of the basic approach I'm using, and if anyone has advice on how to address the remaining tests failures, that would be helpful. Thanks!

P.S. This may also be sufficient to get Intel GPUs working via [HIPCL](https://github.com/cpc/hipcl), though I haven't tested that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ ifdef NVCC_STUBS
     NVCC_LIBS_PATH += -L$(NVCC_STUBS)
 endif
 
-LIBS += -lm -lcudart -lstdc++ -lnvToolsExt -lcufft -lcuda
+LIBS ?= -lm -lcudart -lstdc++ -lnvToolsExt -lcufft -lcuda
 
 
 #############################################################
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ CFLAGS    ?= -fPIC -O3 -funroll-loops -march=native
 CXXFLAGS  ?= $(CFLAGS) -std=c++14
 NVCCFLAGS ?= -std=c++14 -ccbin=$(CXX) -O3 $(NVARCH) -Wno-deprecated-gpu-targets \
 	     --default-stream per-thread -Xcompiler "$(CXXFLAGS)"
+DEVFLAGS  ?= --device-c
 
 # For debugging, tell nvcc to add symbols to host and device code respectively,
 #NVCCFLAGS+= -g -G
@@ -51,7 +52,7 @@ else
 endif
 
 # Common includes
-INC += -I$(CUDA_ROOT)/include -Icontrib/cuda_samples
+INC += -Icontrib/cuda_samples
 
 # NVCC-specific libs
 NVCC_LIBS_PATH += -L$(CUDA_ROOT)/lib64
@@ -102,13 +103,13 @@ CUFINUFFTOBJS_32=$(CUFINUFFTOBJS_64:%.o=%_32.o)
 %_32.o: %.c $(HEADERS)
 	$(CC) -DSINGLE -c $(CFLAGS) $(INC) $< -o $@
 %_32.o: %.cu $(HEADERS)
-	$(NVCC) -DSINGLE --device-c -c $(NVCCFLAGS) $(INC) $< -o $@
+	$(NVCC) -DSINGLE $(DEVFLAGS) -c $(NVCCFLAGS) $(INC) $< -o $@
 %.o: %.cpp $(HEADERS)
 	$(CXX) -c $(CXXFLAGS) $(INC) $< -o $@
 %.o: %.c $(HEADERS)
 	$(CC) -c $(CFLAGS) $(INC) $< -o $@
 %.o: %.cu $(HEADERS)
-	$(NVCC) --device-c -c $(NVCCFLAGS) $(INC) $< -o $@
+	$(NVCC) $(DEVFLAGS) -c $(NVCCFLAGS) $(INC) $< -o $@
 
 default: all
 
--- a/contrib/cuda_hip_wrapper.h
+++ b/contrib/cuda_hip_wrapper.h
@@ -8,17 +8,22 @@
 #include <hipfft.h>
 
 // cuda.h adapters
+#define cudaDeviceReset hipDeviceReset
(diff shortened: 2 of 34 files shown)


elliottslaughter:
Update: I upgraded to ROCm/HIP 4.5.0 (from 4.2.0) and it resolved the issue with symbols. The remaining tests that fail seem to be some sort of a numerical issue.

ahbarnett:
This seems like a worthy goal. I do not have an AMD GPU to test on - and am frankyl out of my depth - we will need others in the dev team to test!

Your macro-based approach looks good, at least as a way to get HIP started with minimal code changes. It unifies the headers for the two architectures. (Note that you now #include fft and complex headers always where they weren't previously - hope this isn't the cause of math errors).
Certainly better than a search-and-replace recommended at blogs like
https://www.lumi-supercomputer.eu/preparing-codes-for-lumi-converting-cuda-applications-to-hip/
!

To complete a PR you would want to add a doc section in README.md or a linked doc file about compiling on HIP. Basically, so everyone knows how to test what you've added.

Try to make the makefile have minimal changes, and a simple flag eg
make check -DHIP

Now, re math, if you can post a single test command line that fails (you say a math-output rather than crash failure?) maybe someone w/ AMD can help us out?

Thanks, Alex

elliottslaughter:
I've been doing some digging and think I've found the root cause, which seems to be related to the host-side code and not to the GPU at all.

My debugging shows that we're not taking the following branch because `MAX_NF` is 0.

https://github.com/flatironinstitute/cufinufft/blob/46a8e8a90049f07e9d35826c80db0ff6bac548bd/contrib/common.cpp#L31

`MAX_NF` meanwhile is defined as `(BIGINT)1e11` as you can see below (note this is `master`, not my branch):

https://github.com/flatironinstitute/cufinufft/blob/46a8e8a90049f07e9d35826c80db0ff6bac548bd/contrib/common.h#L12

And `BIGINT` is defined as `int`, which is 32 bits in most implementations.

https://github.com/flatironinstitute/cufinufft/blob/46a8e8a90049f07e9d35826c80db0ff6bac548bd/contrib/dataTypes.h#L16

To be honest, I'm not sure how this has ever worked. If you compare `1e11` and `1<<32` (which is the largest you can fit in an *unsigned* `int`, so expect about 2x smaller for signed):

```
>>> 1e11
100000000000.0
>>> 1<<32
4294967296
```

You can see that `1e11` clearly overflows. You need to get down to about `1e9` to find something that actually fits in 32 bits.

I have confirmed that changing `1e11` to `1e9` does make the test pass. So it never had anything to do with the GPU in the first place.

I think the potential solutions could be decreasing `MAX_NF` to be `INT_MAX` (or similar), or we could change `BIGINT` to be 64 bits if you really want to be able to represent `1e11` in that number. Either way, it's not really a HIP issue so it should probably be fixed in a separate PR.

MelodyShih:
Hi @elliottslaughter , thanks for reporting this bug and the proposed solutions. This is my mistake. In FINUFFT (the cpu version), `BIGINT` is defined as `int64_t` and everything makes sense. 

I will create a PR to fix it.

elliottslaughter:
For what it's worth, this fix is sufficient to get the entire test suite working on [Spock](https://docs.olcf.ornl.gov/systems/spock_quick_start_guide.html). I'm still seeing issue on [Crusher](https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html) (similar system with newer AMD hardware) that need investigating.

ahbarnett:
Hi Elliot, that's exciting re getting HIP going. Looks like pulling in 1d failed... sorry... but it's a worthwhile part of the code to include. Good luck & thanks for your collaboration on this. -A

elliottslaughter:
@ahbarnett I wanted to follow up to your comment https://github.com/flatironinstitute/cufinufft/pull/134#pullrequestreview-883277905:

> `make check` ... does not have official pass/fail detection; it's a matter of if the tests compile and run (no math check is actually done).

Do the Python tests have such checks? I ask because, so far I've been using `make check` for my AMD/HIP tests, and if they're not checked, it means I haven't actually validated the correctness. If Python does, that may be a path forward though I'll have to work around the lack of PyCUDA on AMD.

janden:
> Do the Python tests have such checks?

They do. I would run `python3 -m pytest python/cufinufft/tests` to verify correctness.

PR #118, opened 2021-11-23
bug fixing for compiling with DEBUG flag; fix input type in spread2d_…

MelodyShih:
Changes:
- `CalcGlobalIdx` is also used in the host code when compile with DEBUG flag on.
- Fix `type` in the default option function in `spread2d_test.cu`.
- Fix opt access in `spread3d_test.cu`.


janden (approved):

janden:
Good catches!

PR #120, opened 2021-11-24
Add type 3 for 3D

AdhocMan:
For a project, we require a type 3 transform on GPU. Since this feature is still missing, I've implemented it for the 3D case. This was mainly done by porting the fiNUFFT implementation for type 3.
The plan setup should also work for 2D, it's just the execution that's missing here. 

This might not be quite ready to merge, since some small changes (such as timers) are still missing. However, it would be helpful to get some feedback from you, to know what's still required to get this merged.


AdhocMan:
It's been a while now, so I'm wondering if anyone would be willing to provide some feedback?

janden:
Apologies for the delay. I'm just getting back from holiday. Hope to look at this later in the week.

contrib/common.cpp
```diff
@@ -95,3 +95,44 @@ void onedim_fseries_kernel(BIGINT nf, FLT *fwkerhalf, SPREAD_OPTS opts)
...
+  legendre_compute_glr(2*q,z,w);        // only half the nodes used, eg on (0,1)
+  for (int n=0;n<q;++n) {
+    z[n] *= J2;                                    // quadr nodes for [0,J/2]
+    f[n] = J2*(FLT)w[n] * evaluate_kernel((FLT)z[n], opts);  // w/ quadr weights
+    //    printf("f[%d] = %.3g\n",n,f[n]);
+  }
+#pragma omp parallel for
+  for (BIGINT j=0;j<nk;++j) {          // loop along output array
+    FLT x = 0.0;                    // register
+    for (int n=0;n<q;++n) x += f[n] * 2*cos(k[j]*z[n]);  // pos & neg freq pair
```
ahbarnett, line 134: Just to catch you up to FINUFFT master, replace   z[n] by (FLT)z[n] here so cos() is single-prec when it should be... improves timing for type 3.

examples/example3d3many.py
```diff
@@ -0,0 +1,71 @@
+"""
+Demonstrate the type 1 NUFFT using cuFINUFFT
```
ahbarnett, line 2: 3d type 3, right?

src/cufinufft.cu
```diff
@@ -1,17 +1,74 @@
...
 #include <iomanip>
+#include <vector>
+#include <complex>
+#include <cstddef>
 #include <math.h>
 #include <helper_cuda.h>
 #include <complex>
 #include <cufft.h>
+// #include <thrust/zip_iterator.h>
+#include <thrust/tuple.h>
```
ahbarnett, line 11: Is thrust a new dependency? or is this just some headers that can be self-contained in the project?

test/cufinufft3d3_test.cu
```diff
@@ -0,0 +1,228 @@
...
+#include "../contrib/utils.h"
+
+using namespace std;
+
+int main(int argc, char* argv[])
+{
+	int N, M;
+	if (argc<3) {
+		fprintf(stderr,
+			"Usage: cufinufft3d2_test method N1 N2 N3 [M [tol]]\n"
```
ahbarnett, line 19: 3d3 not 3d2  in usage string

ahbarnett:
Dear Simon,
This is an impressive and ambitious feature addition!
This required understanding the inner workings of FINUFFT type 3 - not easy - and mashing them up with cuFINUFFT.
We will be delighted to bring it in... given time, since it's nontrivial.
Could I ask what your application is?

I have pulled and tested your C++ tester on an A100, and find that:
1) math seems to check out to required accuracy - congratulations - this is not easy!
2) performance in larger problems is sadly entirely dominated by the trig function evals in setpts.
Eg:
```
(base) ahb@workergpu035 /mnt/home/ahb/numerics/cufinufft> bin/cufinufft3d3_test 1 1e6 1e6 1e-6
[time  ] dummy warmup call to CUFFT	 0.0176 s
[time  ] cufinufft plan:		 2.65e-05 s
[time  ] 	kernel fser (ns=7):	 0.000108 s
[time  ] 	Allocate GPU memory plan 2.61e-05 s
[time  ] 	Copy fwkerhalf1,2 HtoD	 2.66e-05 s
[time  ] 	CUFFT Plan		 0.0139 s
[time  ] 	Allocate GPU memory NUpts0.000238 s
[time  ] 	Setup Subprob properties 0.000648 s
[time  ] 	Allocate GPU memory NUpts0.000263 s
[time  ] 	Setup Subprob properties 0.000784 s
[time  ] cufinufft setNUpts:		 0.67 s
[time  ] 	Initialize fw		 6.14e-06 s
[time  ] 	Spread (1)		 0.0474 s
[time  ] 	Amplify & Copy fktofw	 1.33e-05 s
[time  ] 	CUFFT Exec		 4.4e-05 s
[time  ] 	Unspread (1)		 0.00809 s
[time  ] 	Inner type 2 execution (1)		 0.00819 s
[time  ] 	Deconvolve		 3.89e-05 s
[time  ] cufinufft exec:		 0.0557 s
[time  ] 	Free gpu memory		 0.00417 s
[time  ] 	Free gpu memory		 0.0054 s
[time  ] cufinufft destroy:		 0.00542 s
[Method 1] 1000000 U pts to 1000000 NU pts in 0.731 s:	1.37e+06 NU pts/s
					(exec-only thoughput: 1.8e+07 NU pts/s)
[gpu   ] one targ: rel err in F[500000] is 4.17e-07
```
Even though exec is good (20M pts/sec), overall speed is slower than CPU code on my ryzen laptop
running
```
test/finufft3d_test 1e2 1e2 1e2 1e6 1e-6
```
:(
But this is entirely due to setNUpts.
Solution is to have GPU do the trig funcs, ie reimplement onedim_nuft_kernel on the GPU shared across its threads.
@MelodyShih was due to do this for the CPU function onedim_fseries_kernel, which otherwise is a unnecessary bottleneck in 1d1, 1d2.    I encourage her to try that at a down-moment during thesis write-up :)
But the codes would look v similar.

Other things to keep in mind:

* If you look in my CPU code type-3 tester
https://github.com/flatironinstitute/finufft/blob/69aff52bd5eebb28ea0521bcc303cab1e201715b/test/finufft3d_test.cpp#L153
as the problem size grows I grow the space-bandwidth product by making the box in which the
(s,t,u) freq targs live grow in proportion to N1,N2,N3, which in this case each should be around N^{1/3}.
You'll want to add something like this, since in your tester the FFT is always of size O(1), hardly realistic!  Note cufft is ridiculously fast in the above timings :)
Feel free to bring in that CPU tester. We somehow want to minimize new code (not sure how, other than a joint tester..).

* Since you started, Melody has implemented 1d1 and 1d2, and
I cleaned up the Makefile test tasks to match. You may find it easier to base your Makefile off of that.

* I don't know cupy, but is that something you can apply to the rest of the project or just 3d3?


Ok, that's enough for now! Maybe the rest of the devs have other comments?   Thanks, Alex

test/cufinufft3d3_test.cu
```diff
@@ -0,0 +1,228 @@
...
+	cudaEventRecord(stop);
+	cudaEventSynchronize(stop);
+	cudaEventElapsedTime(&milliseconds, start, stop);
+	totaltime += milliseconds;
+	printf("[time  ] cufinufft destroy:\t\t %.3g s\n", milliseconds/1000);
+
+	checkCudaErrors(cudaMemcpy(fk,d_fk,N*sizeof(CUCPX),cudaMemcpyDeviceToHost));
+
+	printf("[Method %d] %ld U pts to %d NU pts in %.3g s:\t%.3g NU pts/s\n",
+			opts.gpu_method, N ,M,totaltime/1000,M/totaltime*1000);
```
ahbarnett, line 200: For timings in type 3 in CPU code we report (M+N)/totaltime   not M/totaltime.     Ie use the *total* N+M   instead of M each time it appears here. It would be good to make this match the CPU code otherwise we're off by factors of two when M=N, etc...  Thanks, Alex

src/cufinufft.cu
```diff
@@ -1,17 +1,74 @@
...
 
 #include <cufinufft_eitherprec.h>
 #include "cuspreadinterp.h"
 #include "cudeconvolve.h"
 #include "memtransfer.h"
+#include "util_t3.h"
 
 using namespace std;
 
+template <typename T>
```
janden, line 26: I'm not familiar with this construction. Can you explain briefly what's going on? Why is it necessary for the type 3 code?

src/cufinufft.cu
```diff
@@ -127,12 +181,31 @@ This performs:
...
 	  return ier;
 
-	d_plan->dim = dim;
+	SETUP_BINSIZE(type, dim, &d_plan->opts);
+
+	// Setup for type 3 is done when points are set. Only implemented for dim = 3.
+	if(d_plan->type == 3 && d_plan->dim == 3) 
+		return ier;
+
+        // Mult-GPU support: set the CUDA Device ID:
```
janden, line 190: Unindent block.

src/cufinufft.cu
```diff
@@ -315,23 +381,196 @@ Notes: the type FLT means either single or double, matching the
...
+		// pick x, s intervals & shifts & # fine grid pts (nf) in each dim...
+		FLT S1,S2,S3;       // get half-width X, center C, which contains {x_j}...
+		arraywidcen_gpu(M, d_kx, &(d_plan->t3P.X1), &(d_plan->t3P.C1));
+		arraywidcen_gpu(N, d_s, &S1, &(d_plan->t3P.D1));      // same D, S, but for {s_k}
+		set_nhg_type3(S1, d_plan->t3P.X1, d_plan->opts, d_plan->spopts,
+			&(d_plan->nf1), &(d_plan->t3P.h1), &(d_plan->t3P.gam1));  // applies twist i)
+
+		d_plan->t3P.C2 = 0.0;        // their defaults if dim 2 unused, etc
+		d_plan->t3P.D2 = 0.0;
+		if (d_plan->dim>1) {
```
janden, line 402: So it looks like a lot of this code would support dimensions lower than 3. What's missing to get that to work?

src/util_t3.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __UTIL_T3_H__
```
janden, line 1: Maybe just call this `util.h`? Could be a good place to put utility functions (that don't belong in `contrib`) in the future.

janden (changes requested):
Thanks again for putting this together. Will be really nice to have type 3 in cufinufft (also for 1D and 2D soon, hopefully).

I'm not really up on type internals, so I leave that part of the PR to others. Mostly comments/questions there.

test/cufinufft3d3_test.cu
AdhocMan: Makes sense. I've changed it to `M+N`.

python/cufinufft/_cufinufft.py
```diff
@@ -126,14 +126,14 @@ class NufftOpts(ctypes.Structure):
 
 _set_pts = lib.cufinufft_setpts
 _set_pts.argtypes = [
-    c_int, c_void_p, c_void_p, c_void_p, ctypes.c_int, c_double_p,
-    c_double_p, c_double_p, c_void_p]
+    c_int, c_void_p, c_void_p, c_void_p, ctypes.c_int, c_void_p,
+    c_void_p, c_void_p, c_void_p]
```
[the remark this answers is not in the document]
AdhocMan: I believe it caused issues, but I don't remember anymore what exactly.  Theses last few pointers were unused before, but now the types of all pointers match at least.

python/cufinufft/cufinufft.py
```diff
@@ -123,6 +136,21 @@ def __init__(self, nufft_type, modes, n_trans=1, eps=1e-6, isign=None,
         #   we want to keep around for life of instance.
         self.references = []
 
+    @staticmethod
+    def _get_ptr(x):
+        """
+        Return pointer attribute of input. Works with cupy and pycuda arrays.
```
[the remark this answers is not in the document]
AdhocMan: I've reverted the cupy changes and we can look at this separately.

python/cufinufft/cufinufft.py
```diff
@@ -164,38 +192,64 @@ def _plan(self):
...
+        :param t: Array of t points.
+        :param u: Array of u points.
         """
 
         if kx.dtype != self.dtype:
             raise TypeError("cufinufft plan.dtype and "
                             "kx dtypes do not match.")
 
-        if ky and ky.dtype != self.dtype:
+        if ky is not None and ky.dtype != self.dtype:
```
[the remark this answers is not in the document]
AdhocMan: This is cupy related, since cupy arrays cannot be converted to bool.

src/cufinufft.cu
AdhocMan: These structs are used for applying scaling and prephase required for type 3. To avoid having to write kernel launch logic by hand, I'm using `thrust::transform` together with these structs. By using templates, no additional macro naming magic is required to prevent symbol collision between single and double precision object files.

src/cufinufft.cu
AdhocMan: I've written everything with all three dimensions in mind. Supporting 1d and 2d only requires some additional work, replicating the execution in 3d. The `setpts` function should fully support all dimensions already.

src/util_t3.h
AdhocMan: I've moved it to the already existing `include/utils.h` header and added `src/utils.cu` for the kernels.

AdhocMan:
Thank you both for reviewing this.
I'm happy to hear that you could successfully run the tests on your system. 
The application I'm working on is an imaging algorithm for radio astronomy. For us, performance of the execution is more important then `setpts`, since we execute multiple times with the same points. So I did not focus much on accelerating the setup, but good to hear that some work is already being done to help there.

I've rebased my branch on the latest master, so it should now be up to date (including the Makefile, where I also added a single precision test now). The changes required for cupy have been reverted and can be discussed separately.

src/cufinufft.cu
janden: Ok. It would be good to keep these things consistent across the code, however. We have to think about maintenance down the road. If this is only to take care of single vs. double, would you mind using the macro system?

src/cufinufft.cu
janden: Ok cool.

src/util_t3.h
janden: Thanks.

janden:
@AdhocMan Thanks for addressing the changes. There's just one thing left regarding the templates.

@ahbarnett @MelodyShih Are we good otherwise?

AdhocMan:
@janden Thanks for taking another look at this.
Regarding templates, I'd argue that they are far more maintainable, since the compiler automatically takes care of each typed instance. And it looks like, that the only place where such macros are currently defined is in API header files, so adding this implementation detail there would just clutter the interface.
Any chance you could warm up to the idea of using templates? While there may not be instances of defining new templated classes so far, there are certainly many cases of usage through standard library and thrust calls.

janden:
@AdhocMan Sorry for dropping the ball on this.

I would have to look into this further to see whether templates would make sense for us. If we do decide to go down that route, however, it would be for all the different types (and dimensions). Having different approaches in different parts of the codebase would make things more difficult to maintain. I suggest you make an issue with some examples of how things would look like and we can continue the discussion there.

For now, would you be willing to convert this to use macros? Otherwise I can take a stab.

AdhocMan:
I likely won't have the time to look into the conversion, but I'd say templates are certainly easier to maintain, since the compiler does all the work of creating individual symbol names for each type.
However, feel free to change the templated code to macros here.

matthieumeo:
Hello, 

Thank you very much for the nice work. I am developing a Python computational imaging library and I would like to offer support for NUFFTs of Type 3 on GPUs (for MRI and RA imaging). I am therefore very much looking forward to this PR being merged. @janden et al., do you guys have a tentative timeline? When can we expect this work to be available as part as cuFINUFFT. For information, I am already relying on the FINUFFT library for CPU NUFFTs, so cuFINUFFT is a natural choice for me to support GPU NUFFTs. 

Thanks a lot for your time and work.

janden:
> I'd say templates are certainly easier to maintain, since the compiler does all the work of creating individual symbol names for each type.

Ok we'll take this in with templates and look into converting t1 and t2 to templates as well for consistency.

janden:
> @janden et al., do you guys have a tentative timeline?

Currently we're waiting for #142 to be merged and will then try to merge this as soon as possible. I'm not sure what the timeline is for #142 but my guess is that it'll be merged in the next few weeks.

matthieumeo:
Ok thanks, looking forward !

src/cufinufft.cu
janden: Decided to go with templating here. Closing.

PR #121, opened 2021-11-30
Fix NumPy installation

janden:
With the transition from 1.21 to 1.22, NumPy is dropping the binary wheels for manylinux2010 and replacing them with manylinux2014. To make this work for now, we are downgrading NumPy to 1.21 in our requirements.


MelodyShih (approved):
I see. Thanks for the temporary solution.

PR #129, opened 2022-02-04
Attempt to fix Jenkins CI

janden:
Looks like the current issue is that pycuda is compiled against a newer NumPy version (1.22) and then an older version (1.21) installed (following #121). This PR removes the `--system-site-packages` flag for the venv so that it ignores the previously installed (system-wide) packages and installs its own NumPy.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,7 @@ pipeline {
     sh '${PYBIN}/python3 -m venv $HOME'
     sh '''#!/bin/bash -ex
       source $HOME/bin/activate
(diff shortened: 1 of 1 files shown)


janden:
So I'm not 100% sure why replacing `setup.py` with `pip install` makes this works out. Seems that `pip install` should be better about managing dependencies.

ahbarnett (approved):
Looks good to me, and CI works. Thanks for fixing this.  Now we have to decide whether to add `make check` to CI to run Melody's C++ tests. Thoughts?

PR #130, opened 2022-02-04
Update ways to evaluate kernel value when using kerevalmeth=0 (i.e. exp(sqrt())) in interpolation kernels

MelodyShih:
The interpolation kernels when `kernevalmeth = 0` was evaluating the kernel w^d times rather than w*d times as the spreading kernels do. This pull request fixes it. 
p.s. we changed the default value of `kernevalmeth` at some point and these changes should have been done then -- sorry about that.

Here are some performance results before (in the parenthesis) and after the change: 
| GPU | 2D | 3D |
| :---: | :---: | :---: |
| V100 | 1.9 ms (1.8 ms) | 4.0 ms (5.5 ms) |
| RTX8000 | 3.8 ms (6.1 ms)| 24.1 ms (81.0 ms)|

Commands for the tests -- 
    `bin/interp2d_test 1 0 1024 1024 1048576 1e-3`
    `bin/interp3d_test 1 0 128 128 128 2097152 1e-3`

This seems to have a smaller impact on V100. But on RTX8000, we see about 1.6x and 3.4x for the 2D and 3D test case, respectively.

### DIFF
--- a/src/2d/spreadinterp2d.cu
+++ b/src/2d/spreadinterp2d.cu
@@ -518,7 +518,6 @@ void Interp_2d_NUptsdriven(FLT *x, FLT *y, CUCPX *c, CUCPX *fw, int M, const int
 		       int nf1, int nf2, FLT es_c, FLT es_beta, int* idxnupts, int pirange)
 {
 	for(int i=blockDim.x*blockIdx.x+threadIdx.x; i<M; i+=blockDim.x*gridDim.x){
-
 		FLT x_rescaled=RESCALE(x[idxnupts[i]], nf1, pirange);
 		FLT y_rescaled=RESCALE(y[idxnupts[i]], nf2, pirange);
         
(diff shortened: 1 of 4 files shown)


src/2d/spreadinterp2d.cu
```diff
@@ -529,15 +528,21 @@ void Interp_2d_NUptsdriven(FLT *x, FLT *y, CUCPX *c, CUCPX *fw, int M, const int
 		CUCPX cnow;
 		cnow.x = 0.0;
 		cnow.y = 0.0;
+		FLT ker1[MAX_NSPREAD];
+		FLT ker2[MAX_NSPREAD];
+
+		FLT x1=(FLT)xstart-x_rescaled;
+		FLT y1=(FLT)ystart-y_rescaled;
+		eval_kernel_vec(ker1,x1,ns,es_c,es_beta);
+		eval_kernel_vec(ker2,y1,ns,es_c,es_beta);
```
ahbarnett, line 537: in FINUFFT we combine the x and y direction kernels into a single call - might be more efficient (SIMD? not sure matters on GPU).  All allocations are then length 3*MAX_NSPREAD just for convenience in all dims.

src/2d/spreadinterp2d.cu
```diff
@@ -642,15 +648,18 @@ void Interp_2d_Subprob(FLT *x, FLT *y, CUCPX *c, CUCPX *fw, int M, const int ns,
 		xend   = floor(x_rescaled + ns/2.0)-xoffset;
 		yend   = floor(y_rescaled + ns/2.0)-yoffset;
 
+		FLT x1=(FLT)xstart+xoffset-x_rescaled;
+		FLT y1=(FLT)ystart+yoffset-y_rescaled;
+
+		eval_kernel_vec(ker1,x1,ns,es_c,es_beta);
+		eval_kernel_vec(ker2,y1,ns,es_c,es_beta);
```
ahbarnett, line 655: as above

src/3d/spreadinterp3d.cu
```diff
@@ -669,23 +674,27 @@ void Interp_3d_NUptsdriven(FLT *x, FLT *y, FLT *z, CUCPX *c, CUCPX *fw, int M,
 		CUCPX cnow;
 		cnow.x = 0.0;
 		cnow.y = 0.0;
+		FLT x1=(FLT)xstart-x_rescaled;
+		FLT y1=(FLT)ystart-y_rescaled;
```
ahbarnett, line 678: as above etc...

src/3d/spreadinterp3d.cu
```diff
@@ -669,23 +674,27 @@ void Interp_3d_NUptsdriven(FLT *x, FLT *y, FLT *z, CUCPX *c, CUCPX *fw, int M,
...
+		FLT z1=(FLT)zstart-z_rescaled;
+
+		eval_kernel_vec(ker1,x1,ns,es_c,es_beta);
+		eval_kernel_vec(ker2,y1,ns,es_c,es_beta);
```
ahbarnett, line 682: as above, etc...

ahbarnett (approved):
Very glad you caught this. It actually improves the timings from those in our PDSEC paper :)  It's slightly embarrassing I didn't look at this code in detail before...

I make one suggestion which is to stack the kernel evals in to a single call. Don't bother if you think it wouldn't affect SIMD performance (relevant for GPU)?   Probably only for short ns (ie low-acc).

I'll report timings in another comment.  Thanks, Alex

ahbarnett:
Here are timings on an A100:

before:
2d: 1.4ms
3d: 4.8ms

after:
2d: 1.4ms
3d: 2.5ms

So, 2d no effect, 3d twice as fast.
Pretty important.
Cheers, Alex

ahbarnett:
I merged, so you could do the above on another PR or just push to master. Also it would be good to get into the habit of listing improvements in the CHANGELOG...

MelodyShih:
Sounds like a good idea to merge the kernel evaluation calls into one. I will experiment with it. And yes, I will remind myself to update the CHANGELOG in the future. Thank you for updating it with the recent changes.

PR #131, opened 2022-02-06
modify MAX_NF and temporarily remove the usage of BIGINT in cufinufft source and test files

MelodyShih:
Thanks elliottslaughter reports the bug of casting an integer (MAX_NF) greater than INT_MAX to `int` (`BIGINT`) in #116.  The casted result will be implementation-defined.  
I adopt the suggestion of making MAX_NF as INT_MAX. 
I think this is an easier fix because I was not careful about using between `BIGINT` and `int`...:( If we want to use 64 bits integer, more changes need to be involved. For consistency, I removed the usage of `BIGINT` in cufinufft source codes and use `int` everywhere.

@ahbarnett, please let me know what do you think about this. Thanks!

### DIFF
--- a/contrib/common.h
+++ b/contrib/common.h
@@ -9,7 +9,9 @@
 // constants needed within common
 #define MAX_NQUAD 100              // max number of positive quadr nodes
 // increase this if you need >1TB RAM...
-#define MAX_NF    (BIGINT)1e11     // too big to ever succeed (next235 takes 1s)
+#define MAX_NF    (BIGINT)INT_MAX  // too big to ever succeed (next235 takes 1s)
+                                   // for cufinufft, BIGINT is int and MAX_NF is set to
+                                   // max value for int.
 
 struct cufinufft_opts;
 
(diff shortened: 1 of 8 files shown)


contrib/common.h
```diff
@@ -9,7 +9,9 @@
 // constants needed within common
 #define MAX_NQUAD 100              // max number of positive quadr nodes
 // increase this if you need >1TB RAM...
-#define MAX_NF    (BIGINT)1e11     // too big to ever succeed (next235 takes 1s)
+#define MAX_NF    (BIGINT)INT_MAX  // too big to ever succeed (next235 takes 1s)
```
ahbarnett, line 12: Please remove my comment, and replace with something relevant, eg: in this GPU version we only allow array sizes up to 2^31 which about 2e9.

ahbarnett (approved):
Great. Thanks for this fix. Best wishes, Alex

contrib/common.h
ahbarnett: dont' worry, just did it -A

PR #132, opened 2022-02-12
GPU Implementation of 2nd half of onedim_fseries_kernel()

MelodyShih:
The pull request adds a GPU implementation of the 2nd half of function `onedim_fseries_kernel()` and its relative test code and scripts (see [fseries_kernel_test.cu](https://github.com/MelodyShih/cufinufft/blob/onedim_fseries_kernel/test/fseries_kernel_test.cu) and [fseriesperf.sh](https://github.com/MelodyShih/cufinufft/blob/onedim_fseries_kernel/test/fseriesperf.sh)). 

The timing results (tol=1e-6) on a V100 GPU and a Intel Xeon Platinum 8268 CPU shows that it gives a speedup ranges from 0.8x to 27.3x:

According to this timing, I add a heuristic in `src/cufinufft.cu` to switch between the CPU version and the GPU version basing on `nf1`, `nf2` and `nf3`.

ps. the pull request also includes minor updates in the print statement of interpolation kernels.

### DIFF
--- a/contrib/common.cpp
+++ b/contrib/common.cpp
@@ -59,44 +59,29 @@ void onedim_fseries_kernel(BIGINT nf, FLT *fwkerhalf, SPREAD_OPTS opts)
   sampled kernel, not quite the same object.
 
   Barnett 2/7/17. openmp (since slow vs fftw in 1D large-N case) 3/3/18
+  Melody 2/20/22 separate into precomp & comp functions defined below.
  */
 {
-  FLT J2 = opts.nspread/2.0;            // J/2, half-width of ker z-support
-  // # quadr nodes in z (from 0 to J/2; reflections will be added)...
-  int q=(int)(2 + 3.0*J2);  // not sure why so large? cannot exceed MAX_NQUAD
-  FLT f[MAX_NQUAD]; double z[2*MAX_NQUAD],w[2*MAX_NQUAD];
-  legendre_compute_glr(2*q,z,w);        // only half the nodes used, eg on (0,1)
+  FLT f[MAX_NQUAD];
   dcomplex a[MAX_NQUAD];
-  for (int n=0;n<q;++n) {               // set up nodes z_n and vals f_n
-    z[n] *= J2;                         // rescale nodes
-    f[n] = J2*(FLT)w[n] * evaluate_kernel((FLT)z[n], opts); // vals & quadr wei
-    a[n] = exp(2*PI*IMA*(FLT)(nf/2-z[n])/(FLT)nf);  // phase winding rates
-  }
-  BIGINT nout=nf/2+1;                   // how many values we're writing to
-  int nt = MIN(nout,MY_OMP_GET_MAX_THREADS());  // how many chunks
-  std::vector<BIGINT> brk(nt+1);        // start indices for each thread
-  for (int t=0; t<=nt; ++t)             // split nout mode indices btw threads
-    brk[t] = (BIGINT)(0.5 + nout*t/(double)nt);
-#pragma omp parallel
-  {
-    int t = MY_OMP_GET_THREAD_NUM();
-    if (t<nt) {                         // could be nt < actual # threads
-      dcomplex aj[MAX_NQUAD];           // phase rotator for this thread
-      for (int n=0;n<q;++n)
-	aj[n] = pow(a[n],(FLT)brk[t]);       // init phase factors for chunk
-      for (BIGINT j=brk[t];j<brk[t+1];++j) {       // loop along output array
-	FLT x = 0.0;                       // accumulator for answer at this j
-	for (int n=0;n<q;++n) {
-	  x += f[n] * 2*real(aj[n]);       // include the negative freq
-	  aj[n] *= a[n];                   // wind the phases
-	}
-	fwkerhalf[j] = x;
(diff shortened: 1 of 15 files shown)


contrib/common.cpp
```diff
@@ -95,3 +95,47 @@ void onedim_fseries_kernel(BIGINT nf, FLT *fwkerhalf, SPREAD_OPTS opts)
...
+  double z[2*MAX_NQUAD],w[2*MAX_NQUAD];
+  legendre_compute_glr(2*q,z,w);        // only half the nodes used, eg on (0,1)
+  for (int n=0;n<q;++n) {               // set up nodes z_n and vals f_n
+    z[n] *= J2;                         // rescale nodes
+    f[n] = J2*(FLT)w[n] * evaluate_kernel((FLT)z[n], opts); // vals & quadr wei
+    a[n] = exp(2*PI*IMA*(FLT)(nf/2-z[n])/(FLT)nf);  // phase winding rates
+  }
+}
+
+#if 0
```
janden, line 113: Might make sense to uncomment this and have `onedim_fseries_kernel` call both functions. This way we make sure that the same code is being used in the full CPU and the CPU/GPU hybrid cases.

Also, the name here could be `onedim_fseries_kernel_compute` or something to avoid the first half/second half terminology.

src/common.cu
```diff
@@ -0,0 +1,58 @@
...
+using namespace std;
+
+/* TODO */
+__global__
+void OnedimFseriesKernel(int nf1, int nf2, int nf3, FLT *f, cuDoubleComplex *a, FLT *fwkerhalf1, FLT *fwkerhalf2, FLT *fwkerhalf3, int ns)
+{
+	FLT J2 = ns/2.0;
+	int q=(int)(2 + 3.0*J2);
+	int nf;
+	//cuDoubleComplex aj[MAX_NQUAD];
```
janden, line 19: Remove.

contrib/common.cpp
```diff
@@ -95,3 +95,47 @@ void onedim_fseries_kernel(BIGINT nf, FLT *fwkerhalf, SPREAD_OPTS opts)
     }
   }
 }
+
+void onedim_fseries_kernel_1sthalf(BIGINT nf, FLT *f, dcomplex *a, SPREAD_OPTS opts)
```
janden, line 99: Could you adapt the doc for `onedim_fseries_kernel` to this function (and the next one) so it's clear what the `f` and `a` intermediate results are?

src/common.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __COMMON_H__
+#define __COMMON_H__ 
+#include <cufinufft_eitherprec.h>
+
+//__global__
+//void OnedimFseriesKernel(int nf, FLT *f, cuDoubleComplex* a, FLT *fwkerhalf, int ns);
```
janden, line 6: Remove (also below)?

src/common.cu
```diff
@@ -0,0 +1,58 @@
...
+
+#include <cuComplex.h>
+#include "precision_independent.h"
+#include "common.h"
+
+using namespace std;
+
+/* TODO */
+__global__
+void OnedimFseriesKernel(int nf1, int nf2, int nf3, FLT *f, cuDoubleComplex *a, FLT *fwkerhalf1, FLT *fwkerhalf2, FLT *fwkerhalf3, int ns)
```
janden, line 14: Is this still `Onedim` if we have up to three dimensions here?

src/precision_independent.cu
```diff
@@ -5,9 +5,18 @@
...
 #include <thrust/device_ptr.h>
 #include <thrust/scan.h>
+#include <cuComplex.h>
 
 #include "precision_independent.h"
 
+/* Auxiliary func to compute power of complex number */
+__device__ RT carg(const CT& z) {return (RT)atan2(ipart(z), rpart(z));} // polar angle
+__device__ RT cabs(const CT& z) {return (RT)cuCabs(z);}
+__device__ CT cpow(const CT& z, const int &n) {
```
janden, line 15: Unused. Remove?

test/fseriesperf.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# basic perf test of spread/interp for 2/3d, single/double
```
janden, line 2: Update description.

janden (changes requested):
Wow this is really impressive! I have some minor comments. The main question here is whether it's worth it to separate the CPU and CPU/GPU cases. From your results, it doesn't seem to be much slower on the GPU for small sizes. Given that this is a precomputation, the impact should be relatively small to always run on the GPU, and the code would be a lot simpler.

MelodyShih:
Hi @janden , thank you for reviewing the codes and the helpful suggestions. I incorporated them accordingly in the latest commit.
I remove the full CPU version of the codes -- agree that keeping one version of the code (CPU/GPU hybrid) is cleaner. Also, for cases that CPU/GPU hybrid version are slower (small nf), the fseries computation is not the bottleneck of the nufft.

If there are other places that requires changes, please let me know, thanks.

janden:
Great! Sorry I dropped the ball on this. Will merge now.

janden (approved):

PR #134, opened 2022-02-15
Run make check in Jenkins.

janden:
Clone of #125. See if creating the PR from the origin repository allows Jenkins to trust the updated Jenkinsfile.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,7 @@ pipeline {
       python3 -m pip install pytest
       python3 -m pytest
     '''
+    sh 'make check'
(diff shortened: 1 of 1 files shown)


ahbarnett (approved):
let's see how this goes. I'm not the Jenkins expert. Let's keep in mind that `make check` takes a few minutes to run, and does not have official pass/fail detection; it's a matter of if the tests compile and run (no math check is actually done).

janden:
> it's a matter of if the tests compile and run (no math check is actually done).

May be something to add. I'll make an issue.

PR #142, opened 2022-07-12
Handle various safety/library compatibility issues

blackwer:
- [ ] Check namespacing for all non-public API calls
- [ ] Hide private headers for installs
- [ ] Check python compatibility
- [ ] Add python to cmake?
- [ ] Integrate tests with ctest
- [ ] Check result of tests, rather than just if they run


janden:
@blackwer What's the status on this? Is there anything I can help with?

blackwer:
I think it's mostly done and just needs testing/approval. Was hoping to get
back to it before my vacation next thursday. I'll try to go through it
today and come up with more concrete thoughts, but any testing/thoughts
from another voice would be very helpful.

janden:
@blackwer Ok no rush. I'll start reading through it and see if anything pops up.

PR #145, opened 2023-01-11
Less ambiguous None check for python ptrs

garrettwrong:
I found that some of the latest installs have been complaining about the base `if ...` checks.  This resolved for me locally so I'm offering it up here as a patch.


janden:
Thanks for spotting this. We've already incorporated it in the new version of the code that's being merged into Finufft, so I'm closing this out.

janden:
Might be a while before the new code is published, so we're going to go ahead and merge this and push out a small release prior to that.

janden (approved):

PR #150, opened 2023-06-10
Final updates v1.3

janden:
Some final changes before the v1.3 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@
 
 cuFINUFFT is a very efficient GPU implementation of the 1-, 2-, and 3-dimensional nonuniform FFT of types 1 and 2, in single and double precision, based on the CPU code [FINUFFT][1].
 
-Note that the Python interface has changed relative to v1.1. Please see [CHANGELOG](CHANGELOG) for details.
-
 cuFINUFFT introduces several algorithmic innovations, including load-balancing, bin-sorting for cache-aware access, and use of fast shared memory.
 Our tests show an acceleration over FINUFFT of up to 10x on modern hardware,
 and up to 100x faster than other established GPU NUFFT codes:
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# cuFINUFFT v1.2
+# cuFINUFFT v1.3
 
 <img align="right" src="docs/logo.png" width="350">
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -19,6 +19,8 @@ at close-to-FFT speeds;
 under the hood, this involves detailed kernel design, custom spreading/interpolation stages, and plain FFTs performed by cuFFT.
 See the [documentation for FINUFFT][3] for a full mathematical description of the transforms and their applications to signal processing, imaging, and scientific computing.
 
+**Note**: We are currently in the process of adapting the cuFINUFFT interface to closer match that of FINUFFT. This will likely break code depending on the current interface once the next release is published. At this point we will publish a migration guide that will detail the exact changes to the interfaces.
```
ahbarnett, line 22: I think we should also say that this is the last release of cufinufft as its own repo, and future access to cufinufft will be as part of the main finufft repo.

python/cufinufft/README.md
```diff
@@ -1,6 +1,6 @@
 # cuFINUFFT v1.2 Python package
```
ahbarnett: 1.3

setup.py
```diff
@@ -29,7 +29,7 @@
 # Python Package Setup
 setup(
     name='cufinufft',
-    version='1.2',
+    version='1.3',
     author='Yu-shuan Melody Shih, Garrett Wright, Joakim Anden, Johannes Blaschke, Alex Barnett',
     author_email='yoyoshih13@gmail.com',
```
ahbarnett, line 34: How about your email here, Joakim? :)

ahbarnett (approved):
This is good, if we add the indication re the shift to finufft repo. (maybe py users don't notice..>) Thanks, Alex

README.md
janden: Yeah I'll put that in the main readme (this one's only for the Python package).

setup.py
janden: Sure why not.

janden: Done (see 1472cb8c96fd7dba21df25336ba292cf07cbbb6a).

README.md
janden: Done (see 21cef9d53428b3b7359660d53737fc6881be730a).

